PR #131, opened 2024-04-24
Patch edit deck and question

AnteVuletic:


app/actions/deck.ts
```diff
@@ -77,11 +77,11 @@ export async function editDeck(data: z.infer<typeof deckSchema>) {
   const validatedFields = deckSchema.safeParse(data);
 
   if (!validatedFields.success) {
-    return false;
+    return { errorMessage: "Validaiton failed" };
```
marvinmarnold, line 80: Can we be more specific about which validations failed?

app/actions/deck.ts
```diff
@@ -190,19 +179,37 @@ export async function editDeck(data: z.infer<typeof deckSchema>) {
...
           },
         });
+
+        const questionOptionPromiseArray = question.questionOptions.map(
+          (qo) => {
+            return tx.questionOption.upsert({
+              create: {
+                isTrue: qo.isTrue,
+                option: qo.option,
+                questionId: question.id ?? 0,
```
marvinmarnold, line 189: In what circumstances will question.id be undefined? Why is defaulting to 0 the right thing to do?

app/actions/question.ts
```diff
@@ -49,26 +49,27 @@ export async function createQuestion(data: z.infer<typeof questionSchema>) {
...
   const validatedFields = questionSchema.safeParse(data);
 
   if (!validatedFields.success) {
-    return false;
+    return { errorMessage: "Validaiton failed" };
   }
 
   if (!data.id) {
-    return false;
+    return { errorMessage: "Id not specified" };
```
marvinmarnold, line 65: What ID? Question ID?

app/actions/question.ts
```diff
@@ -83,35 +84,51 @@ export async function editQuestion(data: z.infer<typeof questionSchema>) {
...
+              ),
+            },
           },
         },
       },
-    },
-  });
+    });
+
+    const questionOptionUpsertPromiseArray = data.questionOptions.map((qo) => {
```
marvinmarnold, line 113: Can you add a comment explaining what this is doing?

app/components/DeckForm/DeckForm.tsx
```diff
@@ -47,8 +50,15 @@ export default function DeckForm({ deck, tags, action }: DeckFormProps) {
 
   return (
     <form
-      onSubmit={handleSubmit((data) => {
-        action({ ...data, tagIds: selectedTagIds, id: deck?.id });
+      onSubmit={handleSubmit(async (data) => {
```
marvinmarnold, line 53: Refactor this into a function.

app/actions/deck.ts
AnteVuletic: The short answer would be that it's not the correct thing to do in this scenario. Question id will never be null here as per filter on line 150, so this is essentially a typescript issue.

I'm open to any suggestions for a better solution.

app/actions/question.ts
AnteVuletic: Given this action is for question entity it felt sufficient to write Id not specified, but I can correct this if you desire.

app/actions/deck.ts
marvinmarnold: Seems like we should check question.id on line 155, just after entering the for loop. If you do `if (!question.id) break` or `if (!question.id) throw new Error...`, then typescript shouldn't complain.

app/actions/question.ts
marvinmarnold: Yes, please be explicit that it's question ID

app/actions/question.ts
AnteVuletic: Inside the code? Which part of the code would need further explaining, as it seems self explanatory to me?

(In case it's not inside the code:
Creating a promise array for question option upsert (insert and update action), then running the async actions all at once instead of in a loop which would block execution per loop iteration)

AnteVuletic:
> Hey @AnteVuletic we'll start doing PR reviews now. Can you please drop a link to the ClickUp issue this addresses?
> 
> What steps can I take to test this PR?

This issue was discussed in previous meets and there is no click up task.

Prerequisites to reproduce:
- Be logged in as a admin account 
- Have questions or decks created (under question or deck tabs)
- Have questions or decks with answers

To reproduce bug:
- Try to edit any question or deck which has answers

app/actions/question.ts
marvinmarnold: Yes, generally please add more comments inside the code itself.

app/actions/deck.ts
AnteVuletic: Would it be more sensible to continue (skip the specific iteration if the question id is undefined) instead of break? Even tho this will never happen as the questions are filtered to have  id's

AnteVuletic: image

It does seem like typescript is not picking up the hint either way

marvinmarnold (approved):

PR #134, opened 2024-04-26
Refactor binary questions to allow arbitrary values for labels

AnteVuletic:
- Description:
This PR addresses the ability to create binary questions with arbitrary values for the option label instead of fixed T/F and Y/N question types.
- ClickUp links: Not applicable
AC 1:
Ability to add binary question using admin panel
AC 2:
Ability to add binary question to deck using admin panel
AC 3:
Ability to answer binary question
AC 4:
Ability to display result for binary question

In order to test this code we need to run the provided seed script inside of deck md, which will verify ability to answer binary questions.
(AC's should be self explanatory, with the caveat that you do need to set yourself as admin user through db query.)

### DIFF
--- a/app/actions/answer.ts
+++ b/app/actions/answer.ts
@@ -62,10 +62,7 @@ export async function saveDeck(request: SaveQuestionRequest[], deckId: number) {
     );
     const isOptionSelected = qo.id === answerForQuestion?.questionOptionId;
 
-    if (
-      qo.question.type === QuestionType.TrueFalse ||
-      qo.question.type === QuestionType.YesNo
-    ) {
+    if (qo.question.type === QuestionType.BinaryQuestion) {
       const isYesOrTrueOption = qo.option === "Yes" || qo.option === "True";
       return {
         percentage: isYesOrTrueOption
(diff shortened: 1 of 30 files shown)


marvinmarnold (commented):
Does this address a ClickUp ticket? 

I'm not clear what `isLeft` means. Can you please clarify?

AnteVuletic:
> ~Does this address a ClickUp ticket?~ Just noticed you said it doesn't.
> 
> I'm not clear what `isLeft` means. Can you please clarify?

Yes.

The motivation is to allow arbitrary values to be given for binary questions. Essentially we used to have two types of binary questions: Yes/No and True/False.

Bartol has asked me to modify the functionality  in a way where we can define any value for the options in the binary questions. Given that this is the case it felt not appropriate to define one option a the true option, as the options can be for example Blue or Red. Given these values are added to the options then  none of these value is particularly representing a true or false statement. Given that in several places we need to differentiate between these two labels in some way we decided to add a flag representing one or the either branch, we came up with `isLeft` flag to differentiate these two values. If you have any better suggestions we can rename this, as long as it's not `isTrue` which would be wrong because of the previous statements.

marvinmarnold (approved):

PR #149, opened 2024-05-24
Add quick profile

AnteVuletic:
- Description:
In this PR I add support for reusable flyouts (will be needed for notifications) and  implementation of the new quick profile.
- ClickUp links:
https://app.clickup.com/t/86bywd0eq

Only transactions currently stores are the ones related to to points.
I do not know how to check  dollar cost of bonk at this moment, if any API is available or there's a way you want it implemented you can tell me @marvinmarnold. I don't know how much of a prio this is, but it seems non trivial.

Given the user logs in he can click on avatar to view his quick profile.


stories/Navbar.stories.tsx
```diff
@@ -10,14 +11,85 @@ const meta = {
     layout: "centered",
   },
   args: {
-    avatarLink: "/",
-    walletLink: "/",
     avatarSrc: AvatarSample.src,
+    address: "BqcQDyZLW1mL14MgfwMsLkifNScDUTmks55R8x1uAwox",
+    bonkBalance: 9999,
+    transactions: [
```
marvinmarnold, line 17: The intention is that we are pulling from the DB. Is there a reason for hardcoded data here?

stories/Navbar.stories.tsx
```diff
@@ -10,14 +11,85 @@ const meta = {
...
+      },
+    ],
   },
   tags: ["autodocs"],
   decorators: (Story) => (
     <div className="w-96">
-      <Story />
+      <DynamicContextProvider
+        settings={{
+          environmentId: "PLACEHOLDER",
```
marvinmarnold, line 88: Why placeholder?

stories/QuickViewProfile.stories.tsx
```diff
@@ -0,0 +1,107 @@
...
+  },
+  args: {
+    address: "BqcQDyZLW1mL14MgfwMsLkifNScDUTmks55R8x1uAwox",
+    bonkAmount: 108184184,
+    dollarAmount: 218.45,
+    onClose: fn(),
+    avatarSrc: AvatarSample.src,
+    isOpen: true,
+    transactions: [
+      {
```
marvinmarnold, line 23: Same question about hardcoded data

marvinmarnold (changes requested):
- I updated the AC after you looked I think. USD value of BONK no longer a requirement. But the BONK amount still is. Let's save this to last and maybe create a new issue for it depending on on the rest.
- After selecting "View tutorial", you see the screen below. It seems repetitive and like it's reused from somewhere else. Can we just throw users directly into the tutorial instead? Screenshot 2024-05-24 at 3 27 44 PM
- Remove disconnect button from the bottom.
- I don't see any transactions. 
Screenshot 2024-05-24 at 3 36 16 PM

marvinmarnold:
Another request: Please include SOL balance.

stories/Navbar.stories.tsx
AnteVuletic: This is the storybook part. For previewing components

stories/Navbar.stories.tsx
AnteVuletic: Storybook

stories/QuickViewProfile.stories.tsx
AnteVuletic: Storybook

stories/Navbar.stories.tsx
marvinmarnold: Thanks, I'm not familiar with Storybook. Could you please add something to the README explaining how devs should use Storybook in their workflows?

AnteVuletic:
> * I don't see any transactions.


Currently there's no transactions being recorded and we're showing points history. If you answer any question you'll see the points history in here.

> * After selecting "View tutorial", you see the screen below. It seems repetitive and like it's reused from somewhere else. Can we just throw users directly into the tutorial instead? ![Screenshot 2024-05-24 at 3 27 44 PM](https://private-user-


I understand this is not satisfactory UX at the moment. But once this https://github.com/gator-labs/chomp/pull/148 is merged we can fix this issue. 

> * Remove disconnect button from the bottom.

Are you 100% positive if you look at this section: https://www.figma.com/design/jNaJMoRwTJSOBtyLTqWqgC/Gator-UI?node-id=4510-5131&t=eBoSZLoBcxF4YasZ-4, the Disconnect button is there? Can we sync figma perhaps?

stories/Navbar.stories.tsx
AnteVuletic: I added a small explanation of storybook. There detailed documentation on how storybook works, we already discussed internally that we can skip creating stories if we're in a rush.  Also good thing about storybook is that it has handy paid features like chromatic, which can be used for visual regression if needed: https://www.chromatic.com/storybook.

marvinmarnold (approved):

PR #152, opened 2024-05-27
feat: genisis-reveal-and-burn

danijel-radomiljac:
- Description: Add check if user has genisis nft to skip transaction.
- ClickUp links: https://app.clickup.com/t/86bywz2jv


app/actions/used-nft-genisis.ts
```diff
@@ -0,0 +1,19 @@
+"use server";
+
+import prisma from "../services/prisma";
+
+export const getUsedGenisisNfts = async () => {
```
marvinmarnold, line 5: Misspelling. Should be `genesis` throughout.

app/providers/RevealProvider.tsx
```diff
@@ -56,18 +66,43 @@ export function RevealContextProvider({ children }: { children: ReactNode }) {
...
+        lastValidBlockHeight: blockhash.lastValidBlockHeight,
+        signature,
+      });
+      setBurnState("burned");
+    } else {
+      await createUsedGenisisNft(genisisNft.id, primaryWallet!.address);
+    }
 
     if (reveal) {
       await reveal.reveal();
```
marvinmarnold, line 108: Reveal should be happening on the backend after server verifies that the tx hash reported by the user is valid and has never been used before. In this context, I'm not talking about making sure the NFT hasn't been used before. We still need to do that, but now I'm talking about the general reveal logic. The tx hash that the user burned with must be valid and only used once.

marvinmarnold (changes requested):
One small change and one big change requested.

app/providers/RevealProvider.tsx
marvinmarnold: We agreed that @bdeak4 will do this as part of work on https://app.clickup.com/t/86bywz1un

marvinmarnold (commented):
The way things are currently setup, the migrations for this already ran in staging. So I should be able to test this before merging. I'll need more time to do that.

PR #159, opened 2024-05-29
Feat - Daily Deck

danijel-radomiljac:
Implemented new daily deck design.

https://app.clickup.com/t/86byy1qnu


app/components/DailyDeckTitle/DailyDeckTitle.tsx
```diff
@@ -1,14 +1,65 @@
...
+        >
+          <div className="flex items-center justify-between">
+            <p className="text-base">
+              You’re viewing a <span className="text-purple">Daily Deck</span>{" "}
+            </p>
+            <div onClick={() => setIsInfoModalOpen(false)}>
+              <CloseIcon width={16} height={16} />
+            </div>
+          </div>
+          <p className="text-sm">
```
marvinmarnold, line 52: Why is this lorem ipsum?

marvinmarnold (changes requested):

app/components/DailyDeckTitle/DailyDeckTitle.tsx
danijel-radomiljac: Because I don't have actual text for this

marvinmarnold (approved):
approving now  but will check closer later

PR #165, opened 2024-05-30
Feat/question page

boranfurkan:
- Description:
- ClickUp links:

In this pr I have resolved the following tickets: 

1-) https://app.clickup.com/t/86byrxfh2
2-) https://app.clickup.com/t/86byrxd0b

I have created single page for revealed questions. It can be accessed like: `BASE_URL/application/answer/reveal/2`. This page checks if user has revealed the question with the id given in the url, if so then it shows the content of the question accordingly, like in the following images: 

image
image

If there is no such question in the database, it throws 404 not found. In case there is a question but user did not reveal it, we redirect them to home page. 

Only one question is for me to ask here is that: 
image
reward amount is hard coded at the moment. If I can be informed how to get the amount, I can change it to dynamic.


marvinmarnold (approved):

danijel-radomiljac:
@boranfurkan resolve conflicts plss

PR #176, opened 2024-06-01
add bonk burn and fix genesis nfts

bdeak4:
- Description: add bonk burn and fix genesis nfts

ClickUp links:
- https://app.clickup.com/t/86bywz1un

What are the acceptance criteria? What are the steps to test that this code is working?
- create question
- answer question
- reveal question
- if you have genesis nft in wallet it will be used for reveal else it will ask you to burn bonk (amount specified on admin)


app/actions/chompResult.ts
```diff
@@ -82,6 +144,57 @@ export async function revealQuestions(questionIds: number[]) {
...
+    const transaction = await CONNECTION.getParsedTransaction(burnTx, {
+      commitment: "confirmed",
+      maxSupportedTransactionVersion: 0,
+    });
+
+    if (!transaction || transaction.meta?.err) {
+      return null;
+    }
+
+    const burnInstruction = transaction.transaction.message.instructions.find(
```
marvinmarnold, line 183: @bdeak4 I haven't looked at your code super close tbh. But could you please provide a brief explanation of:

1. Is this where you are constructing a new burn tx?
2. Can you point me to the line of code that causes the wallet signature popup to appear?
3. How does your approach compare to the burn implementation I wrote for Colisseum? https://github.com/gator-labs/chomp/blob/94cf2ab3ffc8454a4e64cb1f084e6a1ceb840b49/app/utils/solana.ts#L21

bdeak4: no, it is still using genBonkBurnTx function on frontend to generate transaction and trigger wallet popup. this is for checking if transaction is valid on backend before reveal.

PR #181, opened 2024-06-01
Feat/reveal page claim button

boranfurkan:
- Description:
- ClickUp links:

In this pr I have created the rewards claim buttons on reveal page. Task was given in https://app.clickup.com/t/86bz0yjkz 

I have created these three versions of it: 
Screenshot 2024-06-02 at 01 54 34

Screenshot 2024-06-02 at 01 55 19

Screenshot 2024-06-02 at 01 54 05

I am aware the texts do not match 100% with the Figma design however, I had to put something to create different versions of the button. I can change the texts and design accordingly @KZ-Gator's ideas.

### DIFF
--- a/app/components/RewardShow/RewardShow.tsx
+++ b/app/components/RewardShow/RewardShow.tsx
@@ -6,6 +6,7 @@ import BulkIcon from "../Icons/BulkIcon";
 import { InfoIcon } from "../Icons/InfoIcon";
 import Trophy from "../Icons/Trophy";
 import RewardInfoBox from "../InfoBoxes/RevealPage/RewardInfoBox";
+import Pill from "../Pill/Pill";
 
 interface RewardShowProps {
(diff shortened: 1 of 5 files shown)


app/components/RewardShow/RewardShow.tsx
```diff
@@ -27,19 +28,20 @@ const RewardShow = ({ rewardAmount }: RewardShowProps) => {
             <p className="text-[13px] font-normal leading-[17.55px] text-left">
               Claim reward:
```
marvinmarnold, line 29: This should say 'Reward claimed' if already claimed. Is that the case?

boranfurkan: It actually does. If you check the line 54, you will see that.

marvinmarnold (approved):

PR #226, opened 2024-06-05
fix second button

danijel-radomiljac:


app/providers/RevealProvider.tsx
```diff
@@ -182,7 +184,7 @@ export function RevealContextProvider({ children }: { children: ReactNode }) {
             <Button
               variant="white"
               isPill
-              onClick={burnAndReveal}
+              onClick={() => burnAndReveal(!!"useGenesisNft")}
```
marvinmarnold, line 187: Does`!!"useGenesisNft"` evaluate to `true`?

danijel-radomiljac: yea, I wrote it that way to others deves to understand easily what that boolean does...i could also do it with object as prop

marvinmarnold: Ok makes sense. In the future, more comments would generally be appreciated. Could be helpful a place like here particularly. Or an object prop.

But not blocking for now, thanks for explaining.

marvinmarnold (approved):

PR #270, opened 2024-06-12
fix: add extra check in view image component render conditions

boranfurkan:
- Description:
- ClickUp links: https://app.clickup.com/t/86bz3p882

I have added extra checks where we were rendering the `view image` text to open image modal.

### DIFF
--- a/app/components/QuestionAnswerPreview/QuestionAnswerPreview.tsx
+++ b/app/components/QuestionAnswerPreview/QuestionAnswerPreview.tsx
@@ -36,24 +36,26 @@ export default function QuestionAnswerPreview({
         className="absolute bottom-0 left-0 w-full z-10"
       />
       {children}
-      {viewImageSrc && (
-        <div className="flex items-center gap-[6px] mb-1 z-20">
-          <ImageIcon />
-          <button
-            onClick={() => setIsViewImageOpen(true)}
-            className="underline text-white font-sora font-light text-sm"
-          >
-            View Image
-          </button>
-          <Modal
-            isOpen={isViewImageOpen}
-            onClose={() => setIsViewImageOpen(false)}
-            title=""
-          >
-            <img src={viewImageSrc} />
-          </Modal>
-        </div>
-      )}
+      {viewImageSrc &&
+        viewImageSrc.trim() !== "" &&
+        viewImageSrc.startsWith("http") && (
+          <div className="flex items-center gap-[6px] mb-1 z-20">
+            <ImageIcon />
+            <button
+              onClick={() => setIsViewImageOpen(true)}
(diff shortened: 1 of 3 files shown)


app/components/QuestionAnswerPreview/QuestionAnswerPreview.tsx
```diff
@@ -36,24 +36,26 @@ export default function QuestionAnswerPreview({
...
-            onClose={() => setIsViewImageOpen(false)}
-            title=""
-          >
-            <img src={viewImageSrc} />
-          </Modal>
-        </div>
-      )}
+      {viewImageSrc &&
+        viewImageSrc.trim() !== "" &&
+        viewImageSrc.startsWith("http") && (
```
danijel-radomiljac, line 41: Do we have images that does not start with http or images that have this '  ' as their src ?

boranfurkan: I have added check of `startsWith("http")` because we do not upload the images while creating the question, instead we give the link. And those links should always start with http. Also I have made a check for if it is empty string because was not sure if that check was being made while saving it into db.

danijel-radomiljac: If we do not provide a url, NULL will be saved in database. So there is no need for trim condition.

Also I think the root problem is not conditional rendering, previous condition was fine. Problem is that there is no input validation for image in admin panel. So I would add that check on frontend side and backend side.

boranfurkan: Both ways it would work. I added a check in schema and removed extra checks in front end.

danijel-radomiljac: It acutally wouldn't work both ways. If we kept your initial solution I could just add from admin http as an image url. And then you would have the same problem. If we except correct image url from server, then we need to ensure from backend that we are sending right format.

boranfurkan: If you force to break, you can break anything. That panel being used by admins and admins are the ones who need to make sure adding the correct URL.

danijel-radomiljac: I don't agree with you. If we are storing something in our database, it is our responsibility to store it in right format. It does not matter if regular user or admin is triggering that action.

boranfurkan: Is there anything to still change something with this pr? I have added check in validation schema already. Do you want me to do something else?

PR #271, opened 2024-06-12
fix: promise toasts not closing problem

boranfurkan:
- Description:
- ClickUp links: https://app.clickup.com/t/86bz45b8c

In this pr I have resolved the problem with promise toasts not closing when clicked on close icon.


AnteVuletic:
Make sure you don't break reveal all in history page

boranfurkan:
> Make sure you don't break reveal all in history page

Why would it brake the existing toasts? Rendering approach is still same as it was before. Also, why toast would brake the revealing? @AnteVuletic

PR #273, opened 2024-06-13
fix: hard refresh while burning bonk

danijel-radomiljac:
Ticket: https://app.clickup.com/t/86bz37p5g

Reveal all is temporarily commented because we need to test it more, and change the logic there beacuse of new chagnes I made. I will add ticket for that.

@AnteVuletic please take a look at this pr 😄

### DIFF
--- a/app/providers/RevealProvider.tsx
+++ b/app/providers/RevealProvider.tsx
@@ -3,16 +3,21 @@ import { dasUmi } from "@/lib/web3";
 import { useDynamicContext } from "@dynamic-labs/sdk-react-core";
 import { ISolana } from "@dynamic-labs/solana";
 import { publicKey } from "@metaplex-foundation/umi";
+import { ChompResult } from "@prisma/client";
 import classNames from "classnames";
 import {
-  createContext,
   ReactNode,
+  createContext,
   useCallback,
   useContext,
   useEffect,
   useMemo,
   useState,
 } from "react";
+import {
+  createPendingRevealQuestion,
+  getUsersRevealedQuestion,
+} from "../actions/chompResult";
 import { getUsedGenesisNfts } from "../actions/used-nft-genesis";
 import { Button } from "../components/Button/Button";
 import ChompFullScreenLoader from "../components/ChompFullScreenLoader/ChompFullScreenLoader";
@@ -31,6 +36,7 @@ interface RevealContextState {
   openRevealModal: (
     reveal: (burnTx?: string, nftAddress?: string) => Promise<void>,
     amount: number,
+    questionId: number,
     multiple?: boolean,
   ) => void;
   openClaimModal: (claim: () => Promise<void>) => void;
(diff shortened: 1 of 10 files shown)


app/providers/RevealProvider.tsx
```diff
@@ -68,18 +74,28 @@ export function RevealContextProvider({
...
+      if (reveal?.multiple || !primaryWallet?.address || !reveal?.questionId)
+        return;
+
+      const revealedQuestion = await getUsersRevealedQuestion(
+        reveal!.questionId,
+      );
+
+      setPendingRevealQuestion(revealedQuestion);
 
       const usedGenesisNftIds = (await getUsedGenesisNfts()).map(
```
AnteVuletic, line 100: I would put the genesis nft logic in a separate function

marvinmarnold (approved):

app/providers/RevealProvider.tsx
```diff
@@ -68,18 +74,28 @@ export function RevealContextProvider({
...
     async function effect() {
-      if (reveal?.multiple || !primaryWallet?.address) return;
+      if (reveal?.multiple || !primaryWallet?.address || !reveal?.questionId)
+        return;
+
+      const revealedQuestion = await getUsersRevealedQuestion(
+        reveal!.questionId,
+      );
+
+      setPendingRevealQuestion(revealedQuestion);
```
AnteVuletic, line 98: From my understanding if there's a pending transaction then there is no need to check if there's a genesis token as the bonk burn already happened?

app/providers/RevealProvider.tsx
```diff
@@ -139,6 +176,8 @@ export function RevealContextProvider({
           return;
         }
 
+        await createPendingRevealQuestion(reveal!.questionId, signature);
+
         await CONNECTION.confirmTransaction(
```
AnteVuletic, line 181: From my understanding:
https://github.com/solana-labs/solana-web3.js/blob/ce876d99f04d539292abd810acd77a319c52f50d/packages/library-legacy/src/connection.ts#L3913

confirm transaction can fail leaving chomp result insert in db in pending state

app/providers/RevealProvider.tsx
```diff
@@ -110,9 +126,30 @@ export function RevealContextProvider({
   const burnAndReveal = useCallback(
     async (useGenesisNft = false) => {
       let burnTx: string | undefined;
-      if (!useGenesisNft || !genesisNft || reveal?.multiple) {
+
+      if (!!pendingRevealQuestion) {
+        setBurnState("burning");
+        const signature = pendingRevealQuestion.burnTransactionSignature!;
+
+        await CONNECTION.getTransaction(
```
AnteVuletic, line 134: https://github.com/solana-labs/solana-web3.js/blob/ce876d99f04d539292abd810acd77a319c52f50d/packages/library-legacy/src/connection.ts#L4933 

getTransaction here throws a exception if it cannot find the transaction. It would probably be good to clear the chomp result in this scenario so this isn't stuck in pending state.

After the chomp result is cleared it would be good to return to the normal flow and let the user reveal the question.

app/providers/RevealProvider.tsx
```diff
@@ -68,18 +74,28 @@ export function RevealContextProvider({
...
     multiple: boolean;
     reveal: (burnTx?: string, nftAddress?: string) => Promise<void>;
+    questionId: number;
     genesisNft?: string;
   }>();
   const [claim, setClaim] = useState<{
     claim: () => Promise<void>;
   }>();
   const [genesisNft, setGenesisNft] = useState<string | undefined>();
+  const [pendingRevealQuestion, setPendingRevealQuestion] =
```
AnteVuletic, line 84: For me personally naming this "...RevealQuestion" was confusing. I would name it questionChompResult.

PR #280, opened 2024-06-16
Feat/notification center

boranfurkan:
- Description:
- ClickUp links: https://app.clickup.com/t/86byxnvnw

I have implemented the notification centre component. It shows the revealable questions. If there are questions to be revealed red dot on notification bell image appears.


app/components/NotificationCenter/NotificationCenter.tsx
```diff
@@ -0,0 +1,175 @@
...
+        router.refresh();
+      },
+      revealTokenAmount,
+      id,
+    );
+  };
+
+  return (
+    <div className="w-max h-full flex items-center justify-center">
+      <button
```
AnteVuletic, line 76: This HTML should implement already created components for feed preview which are used in deck details and history

boranfurkan: I have no idea what are you talking about. Can you address the problem better?

AnteVuletic: In react you usually do not slap hundred lines of JSX into a component but you try to extrapolate logic into smaller chunks (components). Since the notification sidebar has a feed like the one mentioned in deck details in history you  should use those components to display the cards in this feed as well. With your implementation you're making it easier to have feed cards display different which  is not advisable as they're the same in the design as well and are obviously the same for a reason. You should try harder to reuse and expand functionality of existing components, this principle is called DRY (or Don't Repeat Yourself) you're obviously repeating lines of code and logic here by not using existing components available for you.


I've also implemented stories for these components so it's harder to miss that they're available these are the urls and the previews. 
http://localhost:6006/?path=/docs/cards-feed-question-card--docs
http://localhost:6006/?path=/docs/cards-reveal-feed-question-card--docs

image
image

This screenshot is what you need to implement?

image


If you compare these two why do you decide you need to rewrite this component from scratch in a block of JSX that's just slapped in here?

marvinmarnold: I appreciate your code reviews @AnteVuletic thanks

marvinmarnold:
@boranfurkan looks like a conflict popped up. I'll test this PR after conflict fixed.

boranfurkan:
> @boranfurkan looks like a conflict popped up. I'll test this PR after conflict fixed.

fixing it right now

boranfurkan:
> @boranfurkan looks like a conflict popped up. I'll test this PR after conflict fixed.

Resolved

PR #285, opened 2024-06-18
Feat/deck scripts

boranfurkan:
- Description:
- ClickUp links: https://app.clickup.com/t/86bz6c4nm, https://app.clickup.com/t/86bz7cxrp

In this pr I have created deck scripts for creating both objective and subjective questions. Script creates number of users and make them answer the question randomly for subjective script. On the other hand, you can give extra paramater in the terminal to make  newly created users choose your answer.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,8 @@
     "prisma:generate": "prisma generate",
     "postinstall": "prisma generate",
     "storybook": "storybook dev -p 6006",
-    "dev:subj-question": "ts-node scripts/CreateSubjectiveWithAnswers.ts first",
-    "dev:obj-question": "ts-node scripts/CreateObjectiveWithAnswers.ts first",
+    "dev:subj-question": "ts-node scripts/CreateSubjectiveWithAnswers.ts",
+    "dev:obj-question": "ts-node scripts/CreateObjectiveWithAnswers.ts",
     "build-storybook": "storybook build",
     "test": "jest --watch",
     "gh-test": "jest"
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "prisma:generate": "prisma generate",
     "postinstall": "prisma generate",
     "storybook": "storybook dev -p 6006",
-    "dev:subj-question": "ts-node scripts/CreateSubjectiveWithAnswers.ts",
+    "dev:subj-question": "ts-node scripts/CreateSubjectiveWithAnswers.ts first",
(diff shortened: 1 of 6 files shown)


scripts/CreateObjectiveWithAnswers.ts
```diff
@@ -76,26 +76,47 @@ async function main() {
...
-        selected: true,
-        timeToAnswer: BigInt(Math.floor(Math.random() * 60000)), // Random time to answer within 60 seconds
-      },
-    });
+    for (const option of questionOptions) {
+      await prisma.questionAnswer.create({
+        data: {
+          userId: user.id,
+          questionOptionId: option.id,
+          percentage: Math.floor(Math.random() * 100),
```
marvinmarnold, line 96: The sum of percentages for a user's answer to a single question should add up to 100%. 

If yes is 75%, then the corresponding no response for the user should be 25%. This code gives a random value to both of the users answers.

marvinmarnold (changes requested):

scripts/CreateObjectiveWithAnswers.ts
boranfurkan: Oh fixing it asap. Thanks for letting me know

package.json
```diff
@@ -17,6 +17,8 @@
     "prisma:generate": "prisma generate",
     "postinstall": "prisma generate",
     "storybook": "storybook dev -p 6006",
+    "dev:subj-question": "ts-node scripts/CreateSubjectiveWithAnswers.ts first",
```
marvinmarnold, line 20: Can you remove 'first' from these commands so that we can run `yarn dev:subj-question second`. I realize I was the one who introduced this pattern, thanks for updating.

scripts/CreateObjectiveWithAnswers.ts
```diff
@@ -26,20 +28,18 @@ async function createUsers(count: number) {
...
       isActive: true,
       deckQuestions: {
         create: {
           question: {
             create: {
-              question: "Is the sky blue?",
+              question: tag + ": Is the sky blue?",
               type: QuestionType.BinaryQuestion,
               revealToken: Token.Bonk,
               revealTokenAmount: 100,
```
marvinmarnold, line 45: Github won't let me comment on the line I want to. But lines 50+51 are wrong. isCorrect should be based on optionToSelect, not hardcoded.

PR #289, opened 2024-06-19
second order estimates fix

danijel-radomiljac:
https://app.clickup.com/t/86bz3q6d1

### DIFF
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -233,9 +233,9 @@ export const calculateReward = async (
       second_order_estimate: userAnswer.percentage,
       second_order_mean: (correctOption || calculatedCorrectOption)
         ?.calculatedAveragePercentage,
-      second_order_estimates: question.questionOptions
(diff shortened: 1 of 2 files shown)


app/utils/algo.ts
```diff
@@ -233,9 +233,9 @@ export const calculateReward = async (
       second_order_estimate: userAnswer.percentage,
       second_order_mean: (correctOption || calculatedCorrectOption)
         ?.calculatedAveragePercentage,
-      second_order_estimates: question.questionOptions
-        .find((option) => option.id === userAnswer.questionOptionId)
-        ?.questionAnswers.map((answer) => answer.percentage),
+      second_order_estimates: (
+        correctOption || calculatedCorrectOption
```
marvinmarnold, line 237: Just double checking that if a user answered incorrectly, there score won't show up. Because we create a QuestionAnswer for both options for all users, I have a doubt this is different than the original code.

PR #344, opened 2024-07-17
Feat/deep profile edit

AndrejRadomiljac:
- Description:
- ClickUp links:


danijel-radomiljac:
> @AndrejRadomiljac on the staging link, nothing happens when I click the edit button.
> 
> Screenshot 2024-07-18 at 10 46 26 AM

@marvinmarnold I think it is because event listener is added only to pencil. We can easily add that to the whole button.

app/actions/profile.ts
```diff
@@ -1,31 +1,79 @@
...
-import { z } from "zod";
-import { profileSchema } from "../schemas/profile";
+import { profileSchemaServer } from "../schemas/profile";
 import prisma from "../services/prisma";
+import s3Client from "../services/s3Client";
+import { uploadImageToS3Bucket, validateBucketImage } from "../utils/file";
 import { getJwtPayload } from "./jwt";
 
-export async function updateProfile(data: z.infer<typeof profileSchema>) {
+export async function updateProfile(formData: FormData) {
```
marvinmarnold, line 11: Making comments in random/unrelated places in the code. 

State is frozen after saving. Button should reset. Please also add a confirmation toast that says. "Profile successfully updated."

danijel-radomiljac: Button should me frozen after saving because there is no change.  If there is change then there is nothing new to submit. 

Good catch for toast 👍

app/components/ProfileForm/ProfileForm.tsx
```diff
@@ -1,86 +1,190 @@
...
+            {...register("username")}
           />
         </div>
+        <Button
+          variant="white"
+          type="submit"
+          className="flex items-center !rounded-[32px]"
+          disabled={isSubmitting || !isDirty}
+        >
+          Save changes
```
marvinmarnold, line 137: Should be prompted to confirm exit if trying to leave the page without first saving. 

https://github.com/user-attachments/assets/fe35d2ee-cc29-42f3-98c4-fea34fc74969

app/schemas/profile.ts
```diff
@@ -1,8 +1,46 @@
...
-  lastName: z.string().min(3).nullish(),
-  username: z.string().min(2).nullish(),
+  firstName: z
+    .string()
+    .regex(/^[a-zA-Z0-9]*$/, {
+      message:
+        "Invalid first name. No spaces or special characters are allowed.",
+    })
+    .min(1, { message: "First name must be at least 1 character long." })
+    .max(45, { message: "First name must be at most 45 characters long." })
```
marvinmarnold, line 15: Max should be 20 chars

app/components/ProfileForm/ProfileForm.tsx
```diff
@@ -1,86 +1,190 @@
...
+        </Button>
+        <Button
```
marvinmarnold, line 139: Not a blocker but moving forward, we want to start implementing this behavior for all buttons. Normal, depressed (mouse down), and loading states.

Screenshot 2024-07-18 at 11 16 44 AM

danijel-radomiljac: @marvinmarnold create ticket for that please

app/schemas/profile.ts
AndrejRadomiljac: Yes, good catch. Fixed in the latest commit.

app/actions/profile.ts
AndrejRadomiljac: Added toast message

app/components/ProfileForm/ProfileForm.tsx
danijel-radomiljac: @marvinmarnold I think that is overkill to add that to this type of form. We would need add extra complexity in code to the navigation and footer navigation just because user edited profile and did not save it. This is not like sensitive information in form and user can come back whenever he wants and do the editing. I also checked how other apps handle that, and they do not have any modal when editing is not finished.

PR #351, opened 2024-07-22
Remove validation for lastName as a required field from profileSchema

AndrejRadomiljac:
- Description:
- Removed validation from last name. Right now last name is not required filed.
- 
Screen Shot 2024-07-22 at 18 19 39 PM
Screen Shot 2024-07-22 at 18 20 36 PM

- ClickUp links:

### DIFF
--- a/app/schemas/profile.ts
+++ b/app/schemas/profile.ts
@@ -42,11 +42,13 @@ export const profileSchemaClient = profileSchema
   })
   .refine(
     (data) => {
-      const { firstName } = data;
-      return !!firstName;
+      const { firstName, lastName } = data;
+      // First name is not required if last name is not set.
+      // If last name is set, first name must also be set.
+      return !lastName || (lastName && !!firstName);
     },
     {
-      message: "Firstname is required.",
+      message: "First name must be provided if a last name is set.",
       path: ["firstName"],
     },
   );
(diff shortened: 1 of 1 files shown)


marvinmarnold (changes requested):
This still isn't quite what I'm expecting.

First name is not required. But if last name is set then first name must also be set. You can set first name but not last name.

You can also only set username but leave first name and last name blank.

AndrejRadomiljac:
> This still isn't quite what I'm expecting.
> 
> First name is not required. But if last name is set then first name must also be set. You can set first name but not last name.
> 
> You can also only set username but leave first name and last name blank.

Oh got it. I'll implement it.

AndrejRadomiljac:
> This still isn't quite what I'm expecting.
> 
> First name is not required. But if last name is set then first name must also be set. You can set first name but not last name.
> 
> You can also only set username but leave first name and last name blank.


https://github.com/user-attachments/assets/d68a198f-13d5-4eeb-9776-6adf26101f03

@marvinmarnold I hope, I've achieved desired validation. Ignore error message on video hehe. The validation error message is: "First name must be provided if a last name is set."

marvinmarnold:
Yes logic looks great @AndrejRadomiljac.

Lastly, is it possible to make the error messages say 'first name' instead of 'firstname'? Same question for last name.

AndrejRadomiljac:
> Yes logic looks great @AndrejRadomiljac.
> 
> Lastly, is it possible to make the error messages say 'first name' instead of 'firstname'? Same question for last name.

As I said in the comment this is the error message that I've implemented in the code: "First name must be provided if a last name is set." @marvinmarnold

marvinmarnold (approved):
Thanks I missed your comment about 'first name'. Everything LGTM now.

PR #365, opened 2024-07-29
feat: added sentry

marvinmarnold:


marvinmarnold:
@danijel-radomiljac any objection to adding sentry so we can track errors better?

PR #374, opened 2024-07-31
[PROD-30, PROD-47] feat: reveal questions miniapp

neel-ds:
- Description:
Add reveal questions flow
- ClickUp links:
https://app.clickup.com/t/86bzjfvgh


app/components/Bot/Bot.tsx
```diff
@@ -0,0 +1,387 @@
...
+        value: { blockhash },
+      } = await CONNECTION.getLatestBlockhashAndContext();
+
+      const signer = await primaryWallet!.connector.getSigner<ISolana>();
+
+      const balance = await CONNECTION.getBalance(
+        new PublicKey(primaryWallet!.address),
+      );
+
+      if (balance < 1000) {
```
marvinmarnold, line 98: This should be the actual amount to burn and not hardcoded to 1000.

app/components/Bot/Bot.tsx
```diff
@@ -0,0 +1,387 @@
...
+        throw new Error("Insufficient balance for the transaction");
+      }
+
+      const tx = await genBonkBurnTx(primaryWallet!.address, blockhash, 10);
```
marvinmarnold, line 102: This should not be hardcoded to 10.

app/components/Bot/Bot.tsx
```diff
@@ -0,0 +1,387 @@
...
+  const dataVerification = async (initData: any) => {
+    const options = {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ initData }),
+    };
+    try {
+      const response = await fetch(`/api/validate`, options);
```
marvinmarnold, line 171: There is no point in verifying tg ID on the client side. We can't trust that clients will respect this code.

It also leaves the server functions unprotected. I will clarify the fix in more comments below.

app/components/Bot/Bot.tsx
```diff
@@ -0,0 +1,387 @@
...
+  }, [selectedRevealQuestions, questions]);
+
+  useEffect(() => {
+    setIsLoading(false);
+  }, [isLoggedIn]);
+
+  if (isLoading) {
+    return (
+      <div className="flex min-h-screen items-center justify-center">
+        <svg
```
marvinmarnold, line 228: Normally we define SVGs in their own file. See app/components/Icons

app/queries/bot.ts
```diff
@@ -0,0 +1,41 @@
+"use server";
+
+export const getUserId = async (telegramId: string) => {
```
marvinmarnold, line 3: This is the continuation of my explanation about how we should be validating. 

Currently we are validating from the client then sending requests to the endpoints defined in this file. But these endpoints are unprotected. This means anybody can invoke this endpoint and get information about any user.

Instead, the telegram payload should be supplied here. The payload should be verified server side in this code, not on the client.

marvinmarnold (changes requested):

app/components/Bot/Bot.tsx
neel-ds: This issue will be resolved in Yash's PR. It's not meant to be used here. 
cc: @yash431garg

app/components/Bot/Bot.tsx
neel-ds: Likewise, onBurn function will be used in Yash's PR. Please consider it cc: @yash431garg

app/components/Bot/Bot.tsx
neel-ds: This spinner has been replaced by the Chomp Loader as mentioned cc: @yash431garg

Yihwan:
PROD-30

app/components/Bot/Bot.tsx
yash431garg: Sure, It will removed from here and we will show insufficient balance in the pop up instead. I will update this in my PR.

app/components/Bot/Bot.tsx
yash431garg: Okay, similarly for this as well will update it in the upcoming pr.

app/queries/bot.ts
```diff
@@ -0,0 +1,121 @@
+"use server";
+
+import { IChompUser } from "../interfaces/user";
+import { extractId } from "../utils/telegramId";
+
+export const getUserData = async (telegramId: string) => {
```
marvinmarnold, line 6: I still don't think these methods are properly protected. Malicious browser code can invoke this method by supplying someone else's telegram ID.

Instead, this method should accept the entire signed payload from Telegram. That should be verified here, on the server. Then only then doe we invoke the underlying API endpoint.

marvinmarnold (approved):
I think there are still some security issues. I'll create a follow up issue for that so we can merge this and I can start testing.

PR #375, opened 2024-07-31
has viewed but not submitted fix

danijel-radomiljac:
https://app.clickup.com/t/86bzkrcfx

### DIFF
--- a/app/actions/answer.ts
+++ b/app/actions/answer.ts
@@ -68,7 +68,6 @@ export async function addTutorialPoints(
 }
 
 export async function saveDeck(request: SaveQuestionRequest[], deckId: number) {
-  console.log(request);
   const payload = await getJwtPayload();
   const userId = payload?.sub ?? "";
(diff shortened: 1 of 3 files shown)


marvinmarnold (changes requested):
It looks like the QuestionAnswers are only being saved after the user selects proceed. This means if they exit the tab without clicking the button, they can play again. 

Instead, as soon as the user is displayed the question, it should be marked as not answered. Then updated when the answer comes in.

danijel-radomiljac:
> It looks like the QuestionAnswers are only being saved after the user selects proceed. This means if they exit the tab without clicking the button, they can play again.
> 
> Instead, as soon as the user is displayed the question, it should be marked as not answered. Then updated when the answer comes in.

Okay then if user exit the tab and come back to the app he should be redirected to daily deck and to second question right? Because I was pointing to that problem when I got into this project. If thats the case we need to adjust the flow.

PR #383, opened 2024-08-01
Remove KZ from CODEOWNERS

Yihwan:


.github/CODEOWNERS
```diff
@@ -1,8 +1,13 @@
 # This is a comment.
+
```
Yihwan, line 2: seems like prettier ran on save, lmk if not expected

marvinmarnold: It's intended that it runs on save. Not sure what it changed with the empty lines but should be fine!

marvinmarnold (approved):

PR #397, opened 2024-08-07
add image upload

danijel-radomiljac:
https://app.clickup.com/t/86bzry05p


marvinmarnold (changes requested):
The changes look great. Two changes please:

1. Remove the image URL option for the deck?
2. Add a button to remove the image in case it gets accidentally set

PR #418, opened 2024-08-15
PROD-130 created script for removing duplicates and migration for unique constraint

AndrejRadomiljac:
- Description
 The issue was caused by hasViewedNotSubmitted column. This PR removes all duplicates and legacy records. Also we've added constraint for not being able to have multiple same answer choices for the same question, for the  same user. We should create a ticket for the following case: 
  - when user starts to answer the question and for some reason user decides to refresh the page. User should not be able to access that question again.


danijel-radomiljac (approved):

prisma/migrations/20240815095358_remove_duplicated_question_answer_records_and_add_unique_constraint/migration.sql
```diff
@@ -0,0 +1,26 @@
+DELETE FROM public."QuestionAnswer"
+WHERE "hasViewedButNotSubmitted" = true;
```
marvinmarnold, line 2: Why do we want to delete these? `hasViewedButNotSubmitted` is useful

prisma/migrations/20240815095358_remove_duplicated_question_answer_records_and_add_unique_constraint/migration.sql
```diff
@@ -0,0 +1,26 @@
...
+    ROW_NUMBER() OVER (
+      PARTITION BY "userId",
+      "questionOptionId"
+      ORDER BY "createdAt" DESC,
+        id DESC
+    ) AS row_num
+  FROM public."QuestionAnswer"
+)
+DELETE FROM public."QuestionAnswer"
+WHERE id IN (
```
marvinmarnold, line 14: The idea is that we keep the oldest answer? 

Is it possible that for some users, this answer won't match the corresponding ChompReward?

prisma/migrations/20240815095358_remove_duplicated_question_answer_records_and_add_unique_constraint/migration.sql
danijel-radomiljac: It is useful but it is not implemented in the right way. It needs to be reimplemented. We have a lot of duplications. You can see in a screenrecord that user with id 001261da-8653-4523-86da-025ca24b2499 set percentage for question id of 5 (80%), and record below that is the same question, same user but hasViewedButNotSubmitted is set to true. Does not make sense right? So that data is not relevant. The best way is to do it from scratch with a new ticket. We can talk more about it tomorrow in a meeting. 😄 

https://github.com/user-attachments/assets/a501d86a-79f6-4126-912d-a9ea20a710ce

prisma/migrations/20240815095358_remove_duplicated_question_answer_records_and_add_unique_constraint/migration.sql
danijel-radomiljac: I think that it does not matter which answer we keep because they are the same.

marvinmarnold: I'm pretty nervous about deleting a bunch of data from prod. Are you aware of any side effects if we keep this dirty data? Does anything break?

danijel-radomiljac: If we do not remove them then we can not add constraint to database.

danijel-radomiljac: https://github.com/user-attachments/assets/8fd7ee2c-639d-401a-b9e8-e07bc755b160

If you look better these are pretty much the same records. Delete older or newer one does not make any difference.

AndrejRadomiljac: I agree with @danijel-radomiljac. The whole idea behind the task was finding those duplicates. The difference between those duplicated records is in createdAt timestamp. And it is clear that difference is in milliseconds or even less. Adding constraint will prevent this. In order to add constraint the duplicates should be removed.

marvinmarnold (approved):

PR #419, opened 2024-08-16
PROD-83: Fix/optimize burn fee

yash431garg:

- Description
Added Estimate priority function from helius api to get an approrate fees along with adding compute unit in the transaction to increase the sucess rate of the burn transation with very low fees
Low priority fees, high success rate. 
There is no ui changes

### DIFF
--- a/app/utils/solana.ts
+++ b/app/utils/solana.ts
@@ -9,6 +9,7 @@ import {
   PublicKey,
   Transaction,
 } from "@solana/web3.js";
+import { getRecentPrioritizationFees } from "../queries/getPriorityFeeEstimate";
 
 export const CONNECTION = new Connection(process.env.NEXT_PUBLIC_RPC_URL!);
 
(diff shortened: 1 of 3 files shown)


app/utils/solana.ts
```diff
@@ -21,35 +22,39 @@ export const genBonkBurnTx = async (
...
+
+  const modifyComputeUnits = ComputeBudgetProgram.setComputeUnitLimit({
+    units: 10000,
+  });
+
+  const addPriorityFee = ComputeBudgetProgram.setComputeUnitPrice({
+    microLamports: estimateFee?.result?.priorityFeeLevels?.medium,
+  });
+
+  tx.add(modifyComputeUnits);
```
marvinmarnold, line 55: I read somewhere that it's better to add the fee txs before the actual burn tx. Should we do that here?

app/utils/solana.ts
```diff
@@ -21,35 +22,39 @@ export const genBonkBurnTx = async (
...
   );
 
+  tx.add(burnTxInstruction);
+
   tx.recentBlockhash = blockhash;
   tx.feePayer = burnFromPublic;
 
+  const estimateFee = await getRecentPrioritizationFees(tx);
+
+  const modifyComputeUnits = ComputeBudgetProgram.setComputeUnitLimit({
```
marvinmarnold, line 47: This is twice as high as needed (4810). I think that's actually fine for now, if your compute price is right. But any reason we shouldn't use some lower hardcoded values like 5500?

app/utils/solana.ts
yash431garg: Yea, I also find out that it is good to keep it above the actual burn. Will be updated in next burn.

app/utils/solana.ts
yash431garg: Now, I have added it to a fix number of 5000 with 25% additonal to prevent compute unit budget fail error. 
Screenshot 2024-08-20 at 1 18 02 PM

marvinmarnold (changes requested):
@yash431garg this looks good. But I think we should merge this into `main`. Can you reopen the PR into `main` branch?

marvinmarnold:
Replaced with: https://github.com/gator-labs/chomp/pull/426

PR #427, opened 2024-08-20
PROD-68: Lockdown chomp backend

yash431garg:

- Description
Updated queries to not trust untrusted data and use telegram and dynamic payload instead.
It does not include code changes mostly include enchancements and fixes.


app/queries/bot.ts
```diff
@@ -1,41 +1,38 @@
...
-    method: "GET",
-    headers: {
-      "Content-Type": "application/json",
-      "api-key": process.env.BOT_API_KEY!,
-    },
-  };
-  try {
     const response = await fetch(
-      `${process.env.NEXT_PUBLIC_API_URL}/question/reveal?userId=${userId}`,
+      `${process.env.NEXT_PUBLIC_API_URL}/question/reveal?userId=${userData.id}`,
```
marvinmarnold, line 28: 1. Can you drop a Github link to where this endpoint defined?
2. Why would this need to be a fetch request instead of invoking a server action/function?

app/queries/bot.ts
```diff
@@ -75,80 +86,83 @@ export const processBurnAndClaim = async (
   }
 };
 
-export const verifyPayload = async (
-  initData: any,
+// With dynamic payload
+export const handleCreateUser = async (
+  id: string,
+  tgId: string,
```
marvinmarnold, line 92: id and telegramId are still being supplied by the client. A malicious user could modify these.

app/queries/bot.ts
```diff
@@ -75,80 +86,83 @@ export const processBurnAndClaim = async (
...
   }
 };
 
-export const handleCreateUser = async (
-  id: string,
-  newId: string,
-  tgId: string,
-  address: string,
+// does user already have an account in pwa
+export const isUserExistByEmail = async (
```
marvinmarnold, line 149: Better name: doesUserExistByEmail

app/queries/bot.ts
```diff
@@ -75,80 +86,83 @@ export const processBurnAndClaim = async (
...
   email: string,
-): Promise<IChompUser | null> => {
+): Promise<boolean | null> => {
   const options = {
-    method: "POST",
+    method: "GET",
```
marvinmarnold, line 153: There should be an auth mechanism here too. We don't want to make it possible to know if an arbitrary email has registered.

app/queries/bot.ts
```diff
@@ -1,41 +1,38 @@
...
-
-const getUserData = async (telegramId: string) => {
-  const options = {
-    method: "GET",
-    headers: {
-      "Content-Type": "application/json",
-      "api-key": process.env.BOT_API_KEY!,
-    },
-  };
+export const getRevealQuestionsData = async (authToken: string) => {
```
marvinmarnold, line 6: Not clear what this method does from the name. Generally, more comments would be helpful at the top of API methods.

marvinmarnold (changes requested):
My biggest confusion is why there are so many fetch requests. My impression is that this code was moved from chomp-bot repo, but now fetch requests no longer needed from miniapp to chomp API since they are hosted part of same app. Can be replaced with server actions?

app/queries/bot.ts
yash431garg: Sure, I will add the comment for this function and will keep in mind for future changes.

app/queries/bot.ts
yash431garg: 1. Here is the github link of the above endpoint: [Get Reveal Quesiton](https://github.com/gator-labs/chomp/blob/feat/lockdown-chomp-backend/app/api/question/reveal/route.ts)
2. We use this api in bot as well to fetch the number of question which are ready to reveal that's why it is a fetch request instead of a server action.

app/queries/bot.ts
yash431garg: Sure, I have updated this.

PR #428, opened 2024-08-21
PROD-178 - add-delete-deck-functionality-to-admin-panel

danijel-radomiljac:


app/actions/deck/deck.ts
```diff
@@ -15,6 +15,154 @@ import { formatErrorsToString } from "../../utils/zod";
 import { handleUpsertingQuestionOptionsConcurrently } from "../question/question";
 import { deckInputFactory } from "./factories";
 
+export async function deleteQuestions(questionIds: number[]) {
+  const isAdmin = await getIsUserAdmin();
+
+  if (!isAdmin) {
+    redirect("/application");
```
marvinmarnold, line 22: Is redirecting here sufficient? Shouldn't we `return` to ensure execution doesn't continue?

app/api/deck/route.ts
```diff
@@ -0,0 +1,20 @@
+import { getTotalNumberOfAnswersInDeck } from "@/app/actions/deck/deck";
+
+export async function GET(request: Request) {
+  const url = new URL(request.url);
+  const deckId = Number(url.searchParams.get("deckId"));
+
+  if (isNaN(deckId)) {
```
marvinmarnold, line 7: Can we also protect with `getIsUserAdmin`?

app/actions/deck/deck.ts
danijel-radomiljac: it also causes return

app/api/deck/route.ts
marvinmarnold: Can you please add anyways? I prefer to expose what's minimally necessary.

app/actions/deck/deck.ts
marvinmarnold: Yea, you are right `Invoking the redirect() function throws a NEXT_REDIRECT error and terminates rendering of the route segment in which it was thrown.`

app/api/deck/route.ts
danijel-radomiljac: added ✅

marvinmarnold (approved):

PR #432, opened 2024-08-22
PROD-125 update-burntx-code-to-versioned-style

danijel-radomiljac:


app/utils/solana.ts
```diff
@@ -37,26 +38,27 @@ export const genBonkBurnTx = async (
     DECIMALS,
   );
 
-  tx.recentBlockhash = blockhash;
-  tx.feePayer = burnFromPublic;
-
-  const estimateFee = await getRecentPrioritizationFees(tx);
```
marvinmarnold, line 43: Why do you want to remove this?

danijel-radomiljac: Because i cannot send versionedTx to that fn and after that push that instruction.

danijel-radomiljac: I added fixed fee

marvinmarnold (commented):
I'd rather keep optimal fee calculation, than use versioned txs. Let's close this PR for now.

PR #435, opened 2024-08-23
PROD-183 changing-second-order-slider-does-not-always-enable-chomp-button 

danijel-radomiljac:


stories/PrimarySliderV2.stories.tsx
```diff
@@ -22,5 +22,6 @@ type Story = StoryObj<typeof meta>;
 export const Default: Story = {
   args: {
     value: 10,
+    activateSlider: () => {},
```
marvinmarnold, line 25: @danijel-radomiljac on today's standup, let's talk about how you are using Storybook in your dev process. I'm not using it, so I'm curious.

marvinmarnold (approved):

PR #437, opened 2024-08-26
 PROD-201 - add-ignore-errors-sentr 

danijel-radomiljac:


marvinmarnold (changes requested):
Can you define the errors as a constant and then import into the sentry config files? That way we don't forget to add an error in the future.

marvinmarnold (approved):

PR #438, opened 2024-08-26
PROD-176/sentry-custom-error-spike

danijel-radomiljac:


app/actions/claim.ts
```diff
@@ -237,7 +153,13 @@ export async function claimQuestions(questionIds: number[]) {
     revalidatePath("/application/profile/history");
     return sendTx;
   } catch (e) {
-    console.log("Error while claiming question", e);
+    Sentry.captureException(
+      `User with id: ${payload.sub} is having trouble with claiming questions with next ids: ${questionIds}`,
+      (scope) => {
```
marvinmarnold, line 158: I'm not sure, but I think this may not be necessary if we create a new error type like `class ClaimError extends Error { }`. Combined with my other comment, would lead to something like:

```
const claimError = new ClaimError(`User with id: ${payload.sub} is having trouble with claiming questions with next ids: ${questionIds}`, {cause: e})
Sentry.captureException(claimError)
```

Are there any benefits that setting scopes provides that we would miss with my proposal?

marvinmarnold (changes requested):

app/actions/claim.ts
marvinmarnold: @danijel-radomiljac your recent changes are still missing
1. New error type `ClaimError` so that we can easily identify
2. Pass the original error as the cause to sentry. 

Here is all the code a bit more clearly.
```
class ClaimError extends Error { }
const claimError = new ClaimError(
`User with id: ${payload.sub} is having trouble with claiming questions with next ids: ${questionIds}`, 
{cause: e}
)
Sentry.captureException(claimError)
```

app/actions/claim.ts
```diff
@@ -237,7 +153,12 @@ export async function claimQuestions(questionIds: number[]) {
     revalidatePath("/application/profile/history");
     return sendTx;
   } catch (e) {
-    console.log("Error while claiming question", e);
+    class ClaimError extends Error {}
```
marvinmarnold, line 156: In the future, let's define custom errors in their own file.

marvinmarnold (approved):

PR #441, opened 2024-08-28
PROD-104 Add backup in workflow

neel-ds:
- Description
Scheduled cron job to take backup automatically and store it in S3

Currently enabled for the development branch and can be triggered through GitHub
 
NA

### DIFF
--- a/.github/workflows/backup.yml
+++ b/.github/workflows/backup.yml
@@ -1,0 +1,41 @@
+name: Database Backup and Upload (Development)
+
+on:
+  workflow_dispatch: # Allows manual triggering from GitHub UI
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  backup-and-upload:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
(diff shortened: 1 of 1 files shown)


marvinmarnold (approved):

.github/workflows/backup.yml
```diff
@@ -0,0 +1,43 @@
+name: Database Backup and Upload (Development)
+
+on:
+  schedule:
```
marvinmarnold, line 4: On second thought, let's avoid having to merge this before testing. Can you:

1. Trigger this workflow to run on PRs into main
```
on:
  pull_request:
    branches:
      - main
   schedule:
      - cron: "33 8,20 * * *" # Run at 8:33 AM and 8:33 PM UTC every day
  workflow_dispatch: # Allows manual triggering from GitHub UI
```
2. After testing, update to reference `prod` branch instead of `main` and update all other env variables to point to prod DB.
3. Then I will approve and we can merge.

marvinmarnold (changes requested):

PR #442, opened 2024-08-28
PROD-175/mixpanel-spike 

danijel-radomiljac:
We are showing userID in every event.

In login we are showing wallet connector.

In question reveal we are showing transactionSignature, burnedAmount and questionId.

In claimed question we are showing transactionSignature, claimedAmount and questionId.

Screenshot 2024-08-28 at 13 10 26


Yihwan:
also defer to @marvinmarnold for actual code review

package.json
```diff
@@ -55,6 +55,8 @@
     "dayjs": "^1.11.10",
     "gsap": "^3.12.5",
     "jsonwebtoken": "^9.0.2",
+    "mixpanel": "^0.18.0",
+    "mixpanel-browser": "^2.55.0",
```
marvinmarnold, line 59: Is this actually being used? If not, let's remove.

marvinmarnold (changes requested):

danijel-radomiljac:
> thanks! some questions:
> 
> * is it possible to get city/country data with server side analytics? or is that only available client side?
> * can you send the actual question content in addition to question IDs? primary users of this dashboard will likely be non technical/not be able to lookup questions by ID easily. we might add more question-related content later, but for now i think just ID and the actual question text is sufficient.

yes we can track city/country but we would need to use some external service and probably one of paid versions: https://ipapi.com/product

lib/mixpanel.ts
```diff
@@ -0,0 +1,43 @@
...
+  eventProperties?: Record<string, any>,
+) => {
+  const payload = await getJwtPayload();
+
+  if (!payload) return;
+
+  const userUUID = payload.sub;
+
+  const additionalProperties = {
+    distinct_id: userUUID,
```
marvinmarnold, line 14: @danijel-radomiljac it looks like if we include `ip`, Mixpanel will automatically geolocate

Screenshot 2024-08-29 at 10 34 20 AM

marvinmarnold (changes requested):

app/api/mixpanel/route.ts
```diff
@@ -0,0 +1,22 @@
...
+import Mixpanel from "mixpanel";
+
+const mixpanel = Mixpanel.init(process.env.MIX_PANEL_TOKEN!);
+
+export async function POST(request: Request) {
+  const data = await request.json();
+  try {
+    const { event, properties } = data;
+
+    mixpanel.track(event, properties);
```
marvinmarnold, line 12: @danijel-radomiljac after second look, I actually think it's better to pull IP directly from the request on the server: https://nextjs.org/docs/pages/api-reference/functions/next-request#ip

That way we don't have to do an extra API request client side.

marvinmarnold (changes requested):

marvinmarnold (approved):

PR #443, opened 2024-08-28
PROD-197/create-campaigns-page 

danijel-radomiljac:


app/campaings/page.tsx
```diff
@@ -0,0 +1,30 @@
+import { getCampaigns } from "@/app/queries/campaign";
```
marvinmarnold, line 1: `campaigns` misspelled in `app/campaings/page.tsx` leading to incorrect path

app/campaings/page.tsx
```diff
@@ -0,0 +1,30 @@
...
+import LeaderboardCard from "../components/LeaderboardCard/LeaderboardCard";
+import Main from "../components/Main/Main";
+
+const CampaignLeaderboard = async () => {
+  const campaigns = await getCampaigns();
```
marvinmarnold, line 6: Should only show campaigns that are active and visible

marvinmarnold (changes requested):

app/campaings/page.tsx
danijel-radomiljac: Why active and visible? If thats the case then why do we have two booleans?

marvinmarnold: tbh I'm not sure the exact reason, but it's one of the requirements in Linear. Probably because we are focused on Breakpoint and we don't want to confuse users with campaigns they can't participate in.

In the past, we've talked about wanting to keep showing campaign results (isVisible = true), even if the campaign is no longer ongoing (isActive = false).

@Yihwan

Yihwan: we are working on designs to better distinguish between active/inactive campaigns but those aren't ready yet. so for now, i'd only include active/visible, and then possibly add in inactive later once the designs are ready/if we need to.

marvinmarnold (approved):

PR #449, opened 2024-08-30
PROD-194/deckinfo-card-hi-fi-design-pass & PROD-160/create-deckinfo-card

danijel-radomiljac:
Screenshot 2024-08-30 at 13 44 03
Screenshot 2024-08-30 at 13 44 43

** NOTE **
Need to update buttons when we implement them.


danijel-radomiljac:
Failing because of migrations

PR #450, opened 2024-08-31
PROD-217 Add shadcn component

neel-ds:
- Description
Install shadcn configuration and add button component
Useable button component
NA


components.json
```diff
@@ -0,0 +1,20 @@
...
+    "config": "tailwind.config.ts",
+    "css": "styles/globals.css",
+    "baseColor": "neutral",
+    "cssVariables": false,
+    "prefix": ""
+  },
+  "aliases": {
+    "components": "@/app/components",
+    "utils": "@/lib/utils",
+    "ui": "@/app/components/ui",
```
Yihwan, line 16: to clarify will running shad-cn `npx` install commands install into `app/components` or `app/components/ui`?

Yihwan (approved):
defer actual cr to @marvinmarnold but lgtm! ty for putting this together so quickly.

styles/globals.css
```diff
@@ -21,6 +21,9 @@
   .home-feed-swiper-slide {
     @apply !h-auto;
   }
+  :root {
```
Yihwan, line 24: originally i was thinking maybe we leave the color vars in so we know what to change, but i like your approach of leaving them out since we aren't using them yet 👍

components.json
neel-ds: It will install to `app/components/ui`

styles/globals.css
neel-ds: We can switch to CSS variables whenever needed. I've currently turned it off in the configuration, but it's easy to switch back on. We will just need to replace it with our palette.

PR #451, opened 2024-09-02
PROD-226/add-mixpanel-events-for-claim

danijel-radomiljac:


marvinmarnold (changes requested):
Looks like this is not fully implemented: please send UserId, UserWalletAddress, UserEmail, Username

danijel-radomiljac:
> Looks like this is not fully implemented: please send UserId, UserWalletAddress, UserEmail, Username

I implemented everything except user email, because we are not storing that in db.

marvinmarnold (approved):

PR #471, opened 2024-09-05
PROD-198/hi-fi-designs

danijel-radomiljac:


app/application/campaigns/page.tsx
```diff
@@ -0,0 +1,28 @@
+import CampaignCard from "@/app/components/CampaignCard/CampaignCard";
```
marvinmarnold, line 1: Why did you add /application/campaigns, given that we already have /campaigns?

danijel-radomiljac: Because I thought /campaigns url is for not logged in users.

marvinmarnold: @Yihwan can you confirm?

marvinmarnold: @danijel-radomiljac it looks like `/campaigns` has been removed, but that was the path I was expecting to remain.

Yihwan: yes the initial idea was to have `/campaigns` visible to non logged in users. not sure i fully understand implications of actual URL structure, but i'm pretty agnostic to this and will defer to devs. 

the primary goal here is to provide as much relevant/interesting info to non logged in users instead of showing a "login" cta right away without context. 

so the product requests are: 
* non logged in users can see list of campaigns and can click into campaigndetail
    * on campaigndetail view, non logged in users can still see the individual deck cards, but all the CTAs will be something like "connect wallet" and clicking a deck will start log in flow (sorry forgot about this state and will work with thana to create)
* logged in users can do everything a logged in user can do, plus see 0 decks to answer/reveal (this field can just be hidden for non loggd in users) on the campaigns view. 
* in addition, clicking "Back" on campaigns can be hidden for non logged in users if that is easiest.

Yihwan: ^ i can add the "Connet Wallet" state to CampaignDetails page as a follow up task if that's easier since that was not originally specified

Yihwan (commented):
added small comment about ensuring circlular image sizes in vercel deploy preview; otherwise can merge this from design review after marvin's comments are addressed.

PR #472, opened 2024-09-05
PROD-250: prototype image generation

marvinmarnold:
Implemented the instructions here: https://vercel.com/docs/functions/og-image-generation

Still some follow up work for prod like updating robots: https://vercel.com/docs/functions/og-image-generation#requirements

Test it out like this: https://chomp-git-ma-image-gen-gator-labs.vercel.app/api/og?username=blended3020


marvinmarnold:
@Yihwan you can test out the flow here. See link in description.

PR #480, opened 2024-09-10
PROD-205/spike-refactor-and-make-satoshi-font-sans-and-remove-fontsize

danijel-radomiljac:
@Yihwan @gatorthana


Yihwan (approved):
tbh i think this actually looks pretty good. let's merge and then handle quick one-off font changes in follow up tasks (if needed)

PR #481, opened 2024-09-10
Prod 193/enable navbar dailydeck

yash431garg:
✅ Add Linear ID (e.g., `PROD-123`) to PR title to automatically link issue

- Description
1. User can skip daily deck and it will be shown in the expiring soon section user can access it again from there.
- What are the steps to test that this code is working?
Able to skip daily deck and open again from the expiring soon section.
- Screen shots or recordings for UI changes
Screenshot 2024-09-10 at 8 06 30 PM
Screenshot 2024-09-10 at 8 06 42 PM

PR #488, opened 2024-09-12
PROD-275/question-can-be-answered-that-doesnt-show-up-in-expiring-soon

danijel-radomiljac:
@marvinmarnold Answer page fetched all questions. I filtered to fetch only non answered non daily questions.

### DIFF
--- a/app/queries/question.ts
+++ b/app/queries/question.ts
@@ -140,17 +140,17 @@ export async function getQuestionSchema(
   return questionData;
 }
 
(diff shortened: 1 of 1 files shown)


marvinmarnold (commented):
Thanks for figuring this out @danijel-radomiljac. After this lands https://github.com/gator-labs/chomp/pull/481 we'll start seeing daily deck questions in 'expiring soon'. 

At that point, the current behavior will no longer be confusing. So I think we can close this PR, but I'm very glad you could explain it!

PR #490, opened 2024-09-12
PROD-282/alertdialog-styling-fixesfinalization 

danijel-radomiljac:


gatorthana (changes requested):
Requesting 3 changes!

1. The header text color should be **secondary** (AFADEB)
2. Move the info box (the grey box with info icon) up above the buttons so the buttons remain more clickable at the very bottom
3. The text in buttons should be **bold** instead of **medium**

image

marvinmarnold:
@danijel-radomiljac there are merge conflicts

gatorthana (approved):

PR #493, opened 2024-09-13
Prod 146: Design freindly errors

yash431garg:
- Description
Handle runtime errors with error pages.
1. 404 page on wrong page route
2. 400 client error page on client exception
3. 500 server error page on server exception
Screenshot 2024-09-13 at 12 25 48 PM


yash431garg:
@Yihwan Can you please review the desings here?

app/components/Icons/RefreshIcon.tsx
```diff
@@ -0,0 +1,17 @@
+import { ICON_DEFAULT_WIDTH_HEIGHT, IconProps } from ".";
```
Yihwan, line 1: we no longer need to import icons manually. pls try importing from https://lucide.dev/guide/packages/lucide-react and see if it is easy to get it to look the same as before; if not we can address in separate PR

app/not-found.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </div>
+      </div>
+
+      <div className="flex flex-col items-start text-white mt-2 gap-2">
+        <p className="text-[16px] mb-2">
+          Page not found. Let&apos; s go somewhere else!
+        </p>
+        <p className="text-[14px] mb-2">Optional error description</p>
+        <p className="text-[14px]">
+          (stack trace, other technical info that may be helpful for debugging)
```
Yihwan, line 32: pls remove all placecholder copy. it is helpful for checking design but if no actual copy was provided pls just remove

Yihwan (changes requested):
icon changes are optional if they require more discussion, but we have to remove all placeholder copy before this can be merged

app/not-found.tsx
yash431garg: Sure, I will remove this.

app/components/Icons/RefreshIcon.tsx
yash431garg: I have changed using lucide.

app/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -0,0 +1,78 @@
...
+import Image from "next/image";
+import Link from "next/link";
+import { Button } from "../ui/button";
+import { RefreshCw, ArrowLeft } from 'lucide-react';
+
+function ErrorBoundary({ error, reset }: { error: Error; reset: () => void }) {
+  const router = useRouter();
+
+  useEffect(() => {
+    Sentry.captureException(error);
```
marvinmarnold, line 13: Let's use a custom error type for this so it's easier to track. Like this: https://github.com/gator-labs/chomp/blob/1f14fbbf2335392fff2c3d18af35b3443caa58d3/app/providers/DynamicProvider.tsx#L64

You can call it ErrorBoundaryError or something

marvinmarnold (changes requested):

PR #494, opened 2024-09-13
PROD-264/percent-answers-not-adding-to-100percent 

danijel-radomiljac:


app/queries/answerPercentageQuery.ts
```diff
@@ -33,7 +33,7 @@ export async function answerPercentageQuery(questionOptionIds: number[]) {
                   , 0)
                   * 100) as "firstOrderSelectedAnswerPercentage",
                 (
-                  select round(avg(percentage))
+                  select floor(avg(percentage))
```
marvinmarnold, line 36: Won't this round down both, and possibly result in totals less than 1? Can the math be implemented such that we only calculate one of QuestionAnswer values for one (user, questionOption), then calculate the other as `1 - CALCULATED_FOR_FIRST`?

marvinmarnold (changes requested):

PR #497, opened 2024-09-15
PROD-254/poap-campaign 

danijel-radomiljac:
@Yihwan I added share button, did not add any logic because I do not fully understand what do you mean by "A link to TaskOn that gets loaded from env"


danijel-radomiljac:
I will reopen this pr, need to add link to TaskOn

PR #498, opened 2024-09-15
PROD-277/incorrect-deck-reveal-state 

danijel-radomiljac:
Some older decks do not have activeFromDate so I added createdAt to be read if there is no activeFromDate. Another alternative is to update those decks by hand in db and add them activeFromDate.


Yihwan:
i am going to update this by hand after getting admin panel. will defer to @marvinmarnold on whether we'd still like to merge this (my understanding is that this state should not be possible in the future, but i think the extra redundancy could be nice)

marvinmarnold:
Left a comment: https://linear.app/gator/issue/PROD-277/incorrect-deck-reveal-state#comment-5e7063e8

PR #499, opened 2024-09-15
Fixed deck url

AndrejRadomiljac:

- Description

### DIFF
--- a/app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
+++ b/app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
@@ -43,7 +43,7 @@ export function HomeFeedDeckCard({
 }: HomeFeedDeckCardProps) {
   return (
     <a
-      href={date ? `/daily-deck` : `application/answer/deck/${deckId}`}
+      href={date ? `/daily-deck` : `application/decks/${deckId}`}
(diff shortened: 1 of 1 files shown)


app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
```diff
@@ -43,7 +43,7 @@ export function HomeFeedDeckCard({
 }: HomeFeedDeckCardProps) {
   return (
     <a
-      href={date ? `/daily-deck` : `application/answer/deck/${deckId}`}
+      href={date ? `/daily-deck` : `application/decks/${deckId}`}
```
Yihwan, line 46: this should probably be in a routes constant but that can be follow up dev cleanup task i defer to @marvinmarnold

Yihwan (approved):

app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
AndrejRadomiljac: I could be constant but, since it's only used at this place, I don't think it is needed to separated in this case.
@Yihwan But as you said it's dev cleanup part and discussion.

marvinmarnold (approved):

PR #524, opened 2024-09-18
PROD-296/deck-info-style-changes

yash431garg:
- Description
Updated deckinfo button to shadcn and screen size
Updated ui and functional go back button.
Screenshot 2024-09-17 at 2 39 58 PM

### DIFF
--- a/app/components/PreviewDeckCard/PreviewDeckCard.tsx
+++ b/app/components/PreviewDeckCard/PreviewDeckCard.tsx
@@ -26,11 +26,11 @@ const PreviewDeckCard = ({
         className,
       )}
     >
-      <header className="flex flex-col gap-4">
-        <h1 className="text-[#8C96ED] text-l">{heading}</h1>
+      <div className="flex flex-col gap-4">
+        <h1 className="text-[#8C96ED] text-lg">{heading}</h1>
         {!!description && <p className="text-sm">{description}</p>}
-      </header>
-      <footer className="flex items-center gap-4">
+      </div>
+      <div className="flex items-center gap-4">
         {!!imageUrl && (
           <div className="relative w-[77px] h-[77px]">
             <Image
@@ -43,14 +43,14 @@ const PreviewDeckCard = ({
           </div>
         )}
 
-        <div className="flex flex-col gap-2">
+        <div className="flex flex-col gap-2 h-[333px]">
           {!!footer && <p className="text-sm">{footer}</p>}
           <p className="text-sm">
             Total {totalNumberOfQuestions} card
             {totalNumberOfQuestions > 1 && "s"}
           </p>
         </div>
-      </footer>
+      </div>
 
       <Image
         src={gatorHeadImage}
--- a/app/components/PreviewDeckCard/PreviewDeckCard.tsx
+++ b/app/components/PreviewDeckCard/PreviewDeckCard.tsx
@@ -22,13 +22,13 @@ const PreviewDeckCard = ({
   return (
     <div
       className={cn(
-        "w-full py-6 px-4 bg-gray-700 rounded-lg relative h-full flex flex-col justify-between",
+        "w-full py-6 px-4 bg-gray-700 rounded-lg relative  min-h-[385px] flex flex-col justify-between",
         className,
       )}
     >
-      <div className="flex flex-col gap-4">
-        <h1 className="text-[#8C96ED] text-lg">{heading}</h1>
-        {!!description && <p className="text-sm">{description}</p>}
+      <div className="flex flex-col gap-[20px]">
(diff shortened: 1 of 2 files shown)


app/screens/DeckScreens/DeckScreen.tsx
```diff
@@ -48,17 +48,17 @@ const DeckScreen = ({
...
               Begin deck
+              <CircleArrowRight />
             </Button>
             <Button
-              className="h-50 w-full"
-              onClick={() => router.push("/application")}
+              variant="outline"
+              onClick={() => router.back()}
             >
-              Back to homepage
```
marvinmarnold, line 61: Can't fully test because of: https://linear.app/gator/issue/PROD-323/correct-revealanswer-navigation-from-campaigns

But back button text should change based on how user got here. See instructions here: https://linear.app/gator/issue/PROD-296/deckinfo-card-changes

app/components/PreviewDeckCard/PreviewDeckCard.tsx
```diff
@@ -43,14 +43,14 @@ const PreviewDeckCard = ({
           </div>
         )}
 
-        <div className="flex flex-col gap-2">
+        <div className="flex flex-col gap-2 h-[333px]">
```
marvinmarnold, line 46: Should use min height in case content overflows for some reason

marvinmarnold (changes requested):

app/screens/DeckScreens/DeckScreen.tsx
yash431garg: Yea, I had a discussion regarding back button with @Yihwan.

Ref: https://linear.app/gator/issue/PROD-296/deckinfo-card-changes#comment-7907f42f

Yihwan (changes requested):
defer to @gatorthana for design review. 

just based off the screenshot: 
* buttons look short in height, pls dbl check expected btn height 
* deck heading font size does not match mocks 
* deck footer and number of cards should be grouped together at the bottom. see this example when there is no deck image: https://www.figma.com/design/3pZWm9gRbVMnkAfd5zY7GE/Chomp-UI-Shared-Library?node-id=1867-99&node-type=frame&t=ZjLnuaEXrOd7bezM-0
* miscellaneous spacing and font sizes do not seem to match mocks

gatorthana (changes requested):
Screenshot 2024-09-19 at 1 10 22 AM

Thank you! Requesting a few changs:

1. The title's purple should be AFADEB (secondary)
2. Let's use capitalization convention of first letter caps -- Begin Deck instead of Begin deck.
3. Corner radius of buttons seem less than mocks
4. Progress bar looks thicker than mocks
5. (made this note in figma just now) -- the carrot icon -- this should be a circle, I want to differentiate this from clickable image in the cards.

yash431garg:
Thanks @Yihwan @gatorthana for the changes request. The requested style changes are not in the scope of this ticket and few of the changes are also part of other ticket. It would be better if we can merge those PRs first and create a follow task. So, that we will not end up doing double work.

marvinmarnold:
@yash431garg thanks for calling out scope creep. Can you please link to the other issue(s) that you think are better suited for these comments? A merge conflict has also popped up. Can you resolve?

yash431garg:
> @yash431garg thanks for calling out scope creep. Can you please link to the other issue(s) that you think are better suited for these comments? A merge conflict has also popped up. Can you resolve?

@marvinmarnold the design changes in this ticket will also be relevant here. 
Ref: https://linear.app/gator/issue/PROD-299/update-deckcompletion-screens

marvinmarnold:
@yash431garg there's conflicts now. Will merge once resolved.

Yihwan:
sure i'm open to addressing these changes in a separate task ,bu for future reference if there is a mock associated with a task it is assumed that the view be build according to spec, regardless of whether the task explicitly mentions this.

app/components/PreviewDeckCard/PreviewDeckCard.tsx
```diff
@@ -44,13 +44,13 @@ const PreviewDeckCard = ({
         )}
 
         <div className="flex flex-col gap-2">
-          {!!footer && <p className="text-sm">{footer}</p>}
-          <p className="text-sm">
+          {!!footer && <p className="text-[14px]">{footer}</p>}
```
Yihwan, line 47: why are we using custom font sizes?

yash431garg:
> sure i'm open to addressing these changes in a separate task ,bu for future reference if there is a mock associated with a task it is assumed that the view be build according to spec, regardless of whether the task explicitly mentions this.

Sure @Yihwan will keep in mind.

PR #525, opened 2024-09-18
PROD-295: campaign decks cleanup

yash431garg:
- Description
1. Added route back to campaign on specific campaign page
2. Udated campaign image as fallback image and set default to deck image.
1. Redirecting back to campaigns from /campaigns/{campaignsId}
2. Verifying decks images on /campaigns/{campaignsId}

### DIFF
--- a/app/campaigns/[id]/page.tsx
+++ b/app/campaigns/[id]/page.tsx
@@ -6,6 +6,7 @@ import { ChompResult, Question } from "@prisma/client";
 import Image from "next/image";
 import Link from "next/link";
 import { notFound } from "next/navigation";
+import CampaignsHeader from "@/app/components/CampaignsHeader/CampaignsHeader";
 
 type PageProps = {
   params: {
@@ -20,9 +21,7 @@ const CampaignPage = async ({ params: { id } }: PageProps) => {
 
   return (
     <div className="flex flex-col gap-2 pt-4 overflow-hidden pb-2">
-      <div className="px-4">
-        <BackButton />
-      </div>
+      <CampaignsHeader backAction="campaigns" />
       <div className="p-4 bg-gray-850 flex gap-4">
         <div className="relative w-[100.5px] h-[100.5px]">
           <Image
@@ -62,7 +61,7 @@ const CampaignPage = async ({ params: { id } }: PageProps) => {
             }
             deckQuestions={deck.deckQuestions.map((dq) => dq.question)}
             deckName={deck.deck}
-            imageUrl={campaign.image}
+            imageUrl={deck.imageUrl ? deck.imageUrl : campaign.image}
             revealAtDate={deck.revealAtDate!}
             numberOfQuestionsOptions={
               deck.deckQuestions.flatMap((dq) => dq.question.questionOptions)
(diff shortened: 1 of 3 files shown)


marvinmarnold (changes requested):
Deck does not have image, but does belong to a campaign. But campaign image not showing: https://chomp-6794mhdwn-gator-labs.vercel.app/application/decks/238

yash431garg:
@marvinmarnold If I am not wrong here from the ticket that changes are related to this page /campaigns/1. Showing deck image if there is one otherwise show campign image is added in this file app/campaigns/[id]/page.tsx. 

Screenshot 2024-09-18 at 3 52 57 PM

marvinmarnold:
> @marvinmarnold If I am not wrong here from the ticket that changes are related to this page /campaigns/1. Showing deck image if there is one otherwise show campign image is added in this file app/campaigns/[id]/page.tsx.

@yash431garg yes you are correct I didn't realize the original scope. Would you mind implementing the same logic on the deck info card page /application/decks/ID also?

yash431garg:
> > @marvinmarnold If I am not wrong here from the ticket that changes are related to this page /campaigns/1. Showing deck image if there is one otherwise show campign image is added in this file app/campaigns/[id]/page.tsx.
> 
> @yash431garg yes you are correct I didn't realize the original scope. Would you mind implementing the same logic on the deck info card page /application/decks/ID also?

Sure I will check that out. But I have a doubt images in the campigns are necassary to create it. But in deck it is optional let's say if the deck isn't related to any campaign and deck also doesn't have a image. What we have to do in that case?

marvinmarnold:
> > > @marvinmarnold If I am not wrong here from the ticket that changes are related to this page /campaigns/1. Showing deck image if there is one otherwise show campign image is added in this file app/campaigns/[id]/page.tsx.
> > 
> > 
> > @yash431garg yes you are correct I didn't realize the original scope. Would you mind implementing the same logic on the deck info card page /application/decks/ID also?
> 
> Sure I will check that out. But I have a doubt images in the campigns are necassary to create it. But in deck it is optional let's say if the deck isn't related to any campaign and deck also doesn't have a image. What we have to do in that case?

Good question. In the case where there is no deck image or campaign image, it is fine to not show any image.

yash431garg:
> > > > @marvinmarnold If I am not wrong here from the ticket that changes are related to this page /campaigns/1. Showing deck image if there is one otherwise show campign image is added in this file app/campaigns/[id]/page.tsx.
> > > 
> > > 
> > > @yash431garg yes you are correct I didn't realize the original scope. Would you mind implementing the same logic on the deck info card page /application/decks/ID also?
> > 
> > 
> > Sure I will check that out. But I have a doubt images in the campigns are necassary to create it. But in deck it is optional let's say if the deck isn't related to any campaign and deck also doesn't have a image. What we have to do in that case?
> 
> Good question. In the case where there is no deck image or campaign image, it is fine to not show any image.

@marvinmarnold I have added the changes you mentioned in the PR mentioned below. The reason behined that was there were already changes related to campaign. So, it was easier to handle it there. 

PR: https://github.com/gator-labs/chomp/pull/530

PR #530, opened 2024-09-18
PROD-299/deck-completion-screen

yash431garg:
- Description
Updates in deck completion screen ui
1.Share button functionality 
2.Next Deck functionality 
Next Deck + arrow icon: next deck in campaign (if one exists) else
Next Deck + arrow icon: next any active deck  (if one exists) else
Go home (if no active deck exists)
Screenshot 2024-09-18 at 9 13 55 PM

### DIFF
--- a/app/application/decks/[id]/page.tsx
+++ b/app/application/decks/[id]/page.tsx
@@ -1,5 +1,6 @@
 import ComingSoonDeck from "@/app/components/ComingSoonDeck/ComingSoonDeck";
 import { NoQuestionsCard } from "@/app/components/NoQuestionsCard/NoQuestionsCard";
+import { getCampaignImage } from "@/app/queries/campaign";
 import { getDeckQuestionsForAnswerById } from "@/app/queries/deck";
 import { getNextDeckId } from "@/app/queries/home";
 import DeckScreen from "@/app/screens/DeckScreens/DeckScreen";
@@ -15,6 +16,8 @@ export default async function Page({ params: { id } }: PageProps) {
 
   const campaignId = Number(deck?.campaignId) || null
 
+  const campaignData = await getCampaignImage(campaignId ?? 0)
+
   const nextDeckId = await getNextDeckId(currentDeckId, campaignId);
 
   return (
@@ -26,6 +29,7 @@ export default async function Page({ params: { id } }: PageProps) {
           currentDeckId={deck.id}
           nextDeckId={nextDeckId}
           questions={deck.questions}
+          campaignImage={campaignData?.image ?? ''}
           deckInfo={{
             ...deck.deckInfo!,
             totalNumberOfQuestions: deck.questions.length,
--- a/app/application/decks/[id]/page.tsx
+++ b/app/application/decks/[id]/page.tsx
@@ -1,7 +1,7 @@
 import ComingSoonDeck from "@/app/components/ComingSoonDeck/ComingSoonDeck";
 import { NoQuestionsCard } from "@/app/components/NoQuestionsCard/NoQuestionsCard";
 import { getDeckQuestionsForAnswerById } from "@/app/queries/deck";
-import { getDecksForExpiringSection } from "@/app/queries/home";
+import { getNextDeckId } from "@/app/queries/home";
 import DeckScreen from "@/app/screens/DeckScreens/DeckScreen";
 
 type PageProps = {
@@ -13,17 +13,18 @@ export default async function Page({ params: { id } }: PageProps) {
 
   const deck = await getDeckQuestionsForAnswerById(currentDeckId);
 
-  const decks = await getDecksForExpiringSection();
-  const nextDeck = decks.filter((deck) => deck.id !== currentDeckId)?.[0];
+  const campaignId = Number(deck?.campaignId) || null
+
+  const nextDeckId = await getNextDeckId(currentDeckId, campaignId);
 
   return (
     <div className="h-full pt-3 pb-4">
       {deck?.questions.length === 0 || deck === null ? (
-        <NoQuestionsCard variant={"regular-deck"} nextDeckId={nextDeck?.id} />
+        <NoQuestionsCard variant={"regular-deck"} nextDeckId={nextDeckId} />
       ) : deck?.questions && deck?.questions?.length > 0 && deck?.deckInfo ? (
         <DeckScreen
           currentDeckId={deck.id}
-          nextDeckId={nextDeck?.id}
+          nextDeckId={nextDeckId}
           questions={deck.questions}
           deckInfo={{
             ...deck.deckInfo!,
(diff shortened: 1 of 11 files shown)


app/components/NoQuestionsCard/NoQuestionsCard.tsx
```diff
@@ -25,51 +26,59 @@ export function NoQuestionsCard({
...
-            router.refresh();
-          }}
-        >
-          Home <HalfArrowRightIcon fill="#0D0D0D" />
-        </Button>
-      )}
-    </div>
+
+      <Button
+        onClick={() => window.open('https://rewards.taskon.xyz/campaign/detail/370083059', '_blank')}
```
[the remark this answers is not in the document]
yash431garg: Sure thanks for poining this out.

marvinmarnold:
Also merge conflicts fyi @yash431garg

yash431garg:
> Also merge conflicts fyi @yash431garg

Just resolved them.

marvinmarnold:
@yash431garg please add corresponding env var to preview only for now in Vercel. We will use a different link that is not yet ready in prod.

yash431garg:
> @yash431garg please add corresponding env var to preview only for now in Vercel. We will use a different link that is not yet ready in prod.

Done

marvinmarnold:
@yash431garg build is failing

app/application/decks/[id]/page.tsx
```diff
@@ -12,18 +13,22 @@ export default async function Page({ params: { id } }: PageProps) {
   const currentDeckId = Number(id);
   const deck = await getDeckQuestionsForAnswerById(currentDeckId);
 
-  const decks = await getDecksForExpiringSection();
-  const nextDeck = decks.filter((deck) => deck.id !== currentDeckId)?.[0];
+  const campaignId = Number(deck?.campaignId) || null
+
+  const campaignData = await getCampaignImage(campaignId ?? 0)
```
Comment (marvinmarnold) [inline app/application/decks/[id]/page.tsx:18]: I guess this works but it's a little weird to pass 0 id and forces an extra DB request.

What about

```
const campaignData = campaignId ? await getCampaignImage(campaignId) ? null
```

Yihwan:
i added some design comments in vercel. 

might be out of scope for this PR but can you investigate why going "Back" from the subsequent DeckInfo card deck screen goes home and not back to the previous deck completion screen? i would expect that the Back buton in this case would direct back to the previous deck. and since that deck was completed, it should show the deck completion screen.

app/components/NoQuestionsCard/NoQuestionsCard.tsx
```diff
@@ -25,51 +26,58 @@ export function NoQuestionsCard({
   const router = useRouter();
 
   return (
-    <div className="flex flex-col justify-between h-full w-full">
+    <div className="flex flex-col justify-between h-full w-full gap-[16px]">
```
Yihwan, line 29: why are we using custom px sizes?

app/components/NoQuestionsCard/NoQuestionsCard.tsx
```diff
@@ -25,51 +26,58 @@ export function NoQuestionsCard({
...
-            router.replace("/application");
-            router.refresh();
-          }}
-        >
-          Home <HalfArrowRightIcon fill="#0D0D0D" />
-        </Button>
-      )}
-    </div>
+      <Button
+        onClick={() => window.open(process.env.REWARD_TASKON_URL, '_blank')}
```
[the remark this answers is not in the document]
Yihwan: also probably explains why clicking on deploy preview did not result in any action. it increases overhead ot manage URLs across env so more reason to put in constant

app/components/NoQuestionsCard/constants.tsx
```diff
@@ -18,29 +18,17 @@ export const QUESTION_CARD_CONTENT = {
     backgroundIcon: <TrophyGraphic />,
   },
   "regular-deck": {
-    title: "Nice!",
+    title: "You finished a deck!",
     body: (date?: Date | null) => (
       <>
-        You just chomped through that deck!{" "}
+        Great job Chomping through a deck!
```
Yihwan, line 24: missing space after exclamation point

styles/globals.css
```diff
@@ -28,7 +28,7 @@
 
 @layer components {
   .questions-card {
-    @apply bg-gray-700 min-h-[304px] w-full rounded-xl px-4 py-6 flex flex-col justify-between border-[0.5px] border-gray-500;
+    @apply bg-gray-700 min-h-[385px] w-full rounded-xl pt-[24px] pl-[16px] pr-[16px] flex flex-col justify-between border-[0.5px] border-gray-500;
```
Yihwan, line 31: this can be addressed in follow up tasks if needed, but we should really avoid using bracketed custom values. if we need to change spacing vars to do that then we should make a separate task.

app/components/NoQuestionsCard/NoQuestionsCard.tsx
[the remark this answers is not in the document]
marvinmarnold: In this case, we really do want different urls per environment. We are testing now with a sample taskon link. Mohit hasn't created the prod taskon link yet and we don't want to have to redeploy once it's ready.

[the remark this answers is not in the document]
Yihwan: oh didn't realize we had different taskon urls for env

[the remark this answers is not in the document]
marvinmarnold: >  clicking on deploy preview did not result in any action

@yash431garg are you sure you set the env var? Do you know why @Yihwan saw this behavior. Another reason might be because it opens in a new tab `_blank`. Not sure if some browsers block that.

yash431garg:
> i added some design comments in vercel.
> 
> might be out of scope for this PR but can you investigate why going "Back" from the subsequent DeckInfo card deck screen goes home and not back to the previous deck completion screen? i would expect that the Back buton in this case would direct back to the previous deck. and since that deck was completed, it should show the deck completion screen.

@Yihwan this is already been done in th other PR. You will be can test that in main or in the mentioned PR.

Ref: https://github.com/gator-labs/chomp/pull/524

app/application/decks/[id]/page.tsx
Comment (yash431garg) [reply app/application/decks/[id]/page.tsx:18]: Thanks @marvinmarnold, I agree with your suggestion will keep in mind in future as well.

app/components/NoQuestionsCard/NoQuestionsCard.tsx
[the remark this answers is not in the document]
yash431garg: Just resolved that.

app/components/NoQuestionsCard/NoQuestionsCard.tsx
yash431garg: Agree... I thought of using exact measurements from figma. But just find out that gap-4 does the same thing.

styles/globals.css
yash431garg: I will do it for spacing in the same PR.

Yihwan (approved):

PR #556, opened 2024-09-23
PROD-319/rename-campaigns-to-stacks 

danijel-radomiljac:


Yihwan (commented):
just want to triple check that we have set up redirects and tested those from campaigns -> stacks to avoid any broken link

marvinmarnold (changes requested):
- Can you update this to use a path defined here: https://github.com/gator-labs/chomp/blob/main/lib/urls.ts Generally, lets stop hardcoding paths in the app moving forward.
- It looks like /campaigns has been renamed /stacks. But like yihwan said, it would be good to have redirects from /campaigns -> /stacks as well. That way any old links from Breakpoint will keep working.

danijel-radomiljac:
> * Can you update this to use a path defined here: [`main`/lib/urls.ts](https://github.com/gator-labs/chomp/blob/main/lib/urls.ts?rgh-link-date=2024-09-24T18%3A13%3A29Z) Generally, lets stop hardcoding paths in the app moving forward.
> * It looks like /campaigns has been renamed /stacks. But like yihwan said, it would be good to have redirects from /campaigns -> /stacks as well. That way any old links from Breakpoint will keep working.

Fixed ✅

marvinmarnold (changes requested):
@danijel-radomiljac I think we need to keep the /campaigns routes that were publicly visible and add hard redirects to /stacks. Let me know if I'm missing some other way that people who see an old QR code to app.chomp.games/campaigns will manage to get redirected to /stacks.

This is in addition to the work in this PR which is good. I'm just saying add a step for redirects.

next.config.js
```diff
@@ -24,6 +24,11 @@ const nextConfig = {
         destination: "/application",
         permanent: true,
       },
+      {
+        source: "/application/campaigns",
```
marvinmarnold, line 28: Shouldn't this be `/campaigns` -> `/stacks`?

app/components/CampaignsHeader/CampaignsHeader.tsx
```diff
@@ -1,21 +1,22 @@
...
 import { useRouter } from "next/navigation";
 import { ArrowLeftIcon } from "../Icons/ArrowLeftIcon";
 
 interface CampaignsHeaderProps {
-  heading?: string; 
+  heading?: string;
   backAction: "back" | "campaigns";
 }
 
 const CampaignsHeader: React.FC<CampaignsHeaderProps> = ({
```
marvinmarnold, line 12: Might be worth going through and renaming the classes to, in order to reduce future tech debt. Eg rename to `StacksHeader` and `StacksHeaderProps`. Also all other campaigns classes in other files.

app/components/CampaignLeaderboard/CampaignLeaderboard.tsx
```diff
@@ -8,7 +8,7 @@ const CampaignLeaderboard = async () => {
 
   return (
     <div className="flex flex-col gap-2">
-      <p>Campaign Leaderboard</p>
+      <p>Stack Leaderboard</p>
```
marvinmarnold, line 11: @Yihwan should this be `Stacks` (plural)?

marvinmarnold (changes requested):

marvinmarnold:
@danijel-radomiljac conflicts FYI

marvinmarnold:
Test is failing because migration needs to run

PR #564, opened 2024-09-26
 PROD-348/breakpoint-nft-reveal-issues 

danijel-radomiljac:
@marvinmarnold there was a missing relation from ChompResult table to RevealNft table. Currently I set it as one to many relation, but it needs to be one to one. I did that because default value in one to one relation cannot be null, but I will fix that after we run a script called: reveal-nft.ts

That script will connect records from RevealNft table to ChompResult table.

### DIFF
--- a/app/actions/chompResult.ts
+++ b/app/actions/chompResult.ts
@@ -158,6 +158,20 @@ export async function revealQuestions(
       },
     });
 
+    let revealNftId = null;
+
+    if (isRevealedWithNft) {
+      const revealNft = await prisma.revealNft.create({
+        data: {
+          userId: payload.sub,
+          nftType,
+          nftId: nftAddress,
+        },
+      });
+
+      revealNftId = revealNft.nftId;
+    }
+
     await tx.chompResult.createMany({
       data: [
         ...questionRewards.map((questionReward) => ({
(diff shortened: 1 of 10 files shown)


marvinmarnold (commented):
I don't totally understand all the details about the relations between tables. Some questions:
- Should this be working in preview or does the script need to be run? When I to reveal on staging with address (ip5UyE6cXhy6cmxuaHWzW2pWmkFAZuDg8mjXCSroeA4), I'm not getting an offer to burn using the [NFT](https://solscan.io/token/DWbxC4FDQpgQGMCnjhyoeFkyaXnGy3CU2aKCpoNe2ryU#metadata) that belongs to that wallet.
- At a high level, why were so many changes necessary? I thought we previously had support for multiple collections.

danijel-radomiljac:
> I don't totally understand all the details about the relations between tables. Some questions:
> 
> * Should this be working in preview or does the script need to be run? When I to reveal on staging with address (ip5UyE6cXhy6cmxuaHWzW2pWmkFAZuDg8mjXCSroeA4), I'm not getting an offer to burn using the [NFT](https://solscan.io/token/DWbxC4FDQpgQGMCnjhyoeFkyaXnGy3CU2aKCpoNe2ryU#metadata) that belongs to that wallet.
> * At a high level, why were so many changes necessary? I thought we previously had support for multiple collections.

- This PR has a migration, so it can be tested when it is merge. After it is merged we need to run a script (first localy, then on prod)
- They are necessary because we added this peace of code that was missing in claim action: 
      const tx = await CONNECTION.getTransaction(transactionHash!, {
        commitment: "confirmed",
        maxSupportedTransactionVersion: 0,
      });

      if (!tx) return errorToast("Cannot get transaction");
      
      So this peace of code is causing an error. So I saw there is a missing relation from ChompResults table to a RevealNft.

PR #569, opened 2024-09-27
PROD-267: Color Audit

neel-ds:
### Description
- Make necessary changes in shadcn components and switch to new button wherever possible
- Remove hard coded colors wherever possible to align with purple/gray
- Stick to color primitives and semantic colors
Best to try the whole app and review consistent UI
Screenshot 2024-09-28 at 12 53 33 AM
Screenshot 2024-09-28 at 12 54 02 AM

### DIFF
--- a/app/providers/RevealProvider.tsx
+++ b/app/providers/RevealProvider.tsx
@@ -297,7 +297,7 @@ export function RevealContextProvider({
                       closeRevealModal();
                     }
                   }}
-                  className="absolute top-5 right-6 border-none w-max !p-0 z-50"
+                  className="!absolute !top-3 !right-6 !border-none !w-max !p-0 z-50"
                 >
                   <CloseIcon width={16} height={16} />
                 </Button>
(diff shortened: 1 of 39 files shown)


app/providers/RevealProvider.tsx
```diff
@@ -297,7 +297,7 @@ export function RevealContextProvider({
                       closeRevealModal();
                     }
                   }}
-                  className="absolute top-5 right-6 border-none w-max !p-0 z-50"
+                  className="!absolute !top-3 !right-6 !border-none !w-max !p-0 z-50"
```
yash431garg, line 300: Does adding exclamation mark behind every property is necessary here?

tailwind.config.ts
yash431garg, line 1: Wouldn't using "destructive" instead of "red" be confusing?

app/providers/RevealProvider.tsx
neel-ds: To override the button variant, it's necessary.

tailwind.config.ts
neel-ds: According to Figma color primitives, "destructive" should be used.

yash431garg (approved):

PR #572, opened 2024-10-01
PROD-341: fix login pop up stacks

neel-ds:
- Description
Revert back to #550 and fix loader screen issue on mobile if stack has more decks

The LoginPopup was causing issues on the stack page for mobile screens, so you temporarily commented it out during Breakpoint. This PR resolves the issue by disabling an initial isLoading state in LoginPopup and reverting the original behavior of triggering on deck click.

Open Engima stack in mobile to test the consistency

NA

### DIFF
--- a/app/components/LoginPopUp/LoginPopUp.tsx
+++ b/app/components/LoginPopUp/LoginPopUp.tsx
@@ -21,7 +21,7 @@ type LoginPopUpProps = {
 
 const LoginPopUp = ({ isOpen, onClose, userId, deckId }: LoginPopUpProps) => {
   const { authToken, awaitingSignatureState } = useDynamicContext();
-  const [isLoading, setIsLoading] = useState(true);
+  const [isLoading, setIsLoading] = useState(false);
 
   useEffect(() => {
     if(localStorage.getItem("selectedDeckId")) {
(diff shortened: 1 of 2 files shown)


marvinmarnold (commented):
Could you add more details to your PR description? What was the root cause and how does this change fix it?

neel-ds:
> Could you add more details to your PR description? What was the root cause and how does this change fix it?

The LoginPopup was causing issues on the stack page for mobile screens, so you temporarily commented it out during Breakpoint. This PR resolves the issue by disabling an initial isLoading state in LoginPopup and reverting the original behavior of triggering on deck click.

danijel-radomiljac (approved):

marvinmarnold:
> This PR resolves the issue by disabling an initial isLoading state in LoginPopup and reverting the original behavior of triggering on deck click.

it's not obvious to me why the wrong isLoading initial state was causing the slow behavior on mobile. Can you explain the bug a little more.

neel-ds:
> it's not obvious to me why the wrong isLoading initial state was causing the slow behavior on mobile. Can you explain the bug a little more.

@marvinmarnold  The isLoading state in LoginPopUp was displaying LoaderScreen while fetching authToken and awaitSignatureState from dynamicContext. Since this is inside a useEffect, it could trigger validation multiple times as the context changes. I turned off the initial isLoading state to prevent the loader screen from appearing unnecessarily when the user is on the stack page without interacting with a deck.

https://github.com/gator-labs/chomp/blob/2a9ba15fa908e253de9d7bc4c308f72ffb02c9d5/app/components/LoginPopUp/LoginPopUp.tsx#L61

danijel-radomiljac:
I will close this PR because I will create a new one. I think I found where the problem was.

PR #573, opened 2024-10-01
PROD-318/dont-show-question-to-reveal-with-too-few-answers-for-calculating-the

danijel-radomiljac:


AndrejRadomiljac (dismissed):

app/stacks/[id]/page.tsx
```diff
@@ -38,7 +36,18 @@ const StackPage = async ({ params: { id } }: PageProps) => {
           <h1 className="text-base mb-3">{stack.name}</h1>
           <p className="text-xs mb-6">
             {stack.deck.length} deck{stack.deck.length === 1 ? "" : "s"},{" "}
-            {stack.deck.flatMap((d) => d.deckQuestions).length} cards
+            {
+              stack.deck
```
Comment (marvinmarnold) [inline app/stacks/[id]/page.tsx:40]: Can you refactor this? At the minimum, do this calculation above, before the HTML rendering and store result in var. Then reference variable here. 

Even better would be to define a utility function for this logic.

app/queries/history.ts
```diff
@@ -203,7 +189,8 @@ WHERE
 
   return questions.filter(
     (question) =>
```
marvinmarnold, line 191: Can this repeated logic be refactored?

marvinmarnold (changes requested):
Blocker: Can you add this env var to the README and .env.example

Not blocker: Refactoring

marvinmarnold (approved):

PR #580, opened 2024-10-03
PROD-374/message-relation-stack-does-not-exist 

danijel-radomiljac:


marvinmarnold (approved):
What functionality was breaking because of this?

danijel-radomiljac:
> What functionality was breaking because of this?

It could easily cause application server error and crash the app.

PR #582, opened 2024-10-03
PROD-362 Ignore sentry error

neel-ds:
- Description
`Error: Cannot destructure property 'address' of '(intermediate value)' as it is undefined.` doesn't impact the user experience. It appears to be related to a browser extension, which is beyond our control. Ignoring this sentry error would be better option unless it starts causing any real issues.
Monitor sentry board for similar errors
NA

### DIFF
--- a/app/utils/sentry.ts
+++ b/app/utils/sentry.ts
@@ -4,4 +4,5 @@ export const ignoreErrors = [
   "[PhantomRedirect] getSession called, but required input 'encryptionSecretKey' not found in local storage",
   "User rejected the request.",
   "Talisman extension has not been configured yet. Please continue with onboarding.",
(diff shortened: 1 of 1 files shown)


yash431garg (commented):
Does it comes if user initiate the wallet connection but ends up cancelling it.

neel-ds:
> Does it comes if user initiate the wallet connection but ends up cancelling it?

No, it triggers without any user interaction even before user click on the connect wallet button

yash431garg:
> > Does it comes if user initiate the wallet connection but ends up cancelling it?
> 
> No, it triggers without any user interaction even before user click on the connect wallet button

It occurred around 1.1k for 7 users any thoughts why or which extension may be causing this issue.

yash431garg (approved):

marvinmarnold (approved):

PR #586, opened 2024-10-07
PROD-341/fix-login-popup-from-campaigns

danijel-radomiljac:
The problem was in useEffect, there was a missing guard if (!isOpen) return;

Because without this, useEffect was run in all components and now is run only when popup is opened in specific component.


neel-ds (approved):
LGTM!

PR #592, opened 2024-10-09
PROD-383: Sentry invariant expected app router to be mounted

yash431garg:
- Description
Previously because we had useRouter in our error boundary and global error file catches the error from Root Layout which is out of next context causing this error sentry invariant expected app router to be mounted.
Rendering global-error and error boundary in production properly.
N/A

### DIFF
--- a/app/components/ErrorBoundary/ErrorBoundary.tsx
+++ b/app/components/ErrorBoundary/ErrorBoundary.tsx
@@ -1,6 +1,6 @@
+"use client"
 import * as Sentry from "@sentry/nextjs";
 import { ArrowLeft, RefreshCw } from "lucide-react";
-import { useRouter } from "next-nprogress-bar";
 import Image from "next/image";
 import Link from "next/link";
 import { useEffect } from "react";
@@ -8,8 +8,6 @@ import { Button } from "../ui/button";
 import { HOME_PATH } from "@/lib/urls";
 
 function ErrorBoundary({ error, reset }: { error: Error; reset: () => void }) {
-  const router = useRouter();
-
   useEffect(() => {
     Sentry.captureException(error);
   }, [error]);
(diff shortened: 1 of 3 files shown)


app/components/ErrorBoundary/ErrorBoundary.tsx
```diff
@@ -55,15 +53,14 @@ function ErrorBoundary({ error, reset }: { error: Error; reset: () => void }) {
...
-          <Button
-            variant="outline"
-            className="text-[14px] gap-2"
-          >
-            <ArrowLeft />
-            Return home
-          </Button>
-        </Link>
+        <Button
+          variant="outline"
```
marvinmarnold, line 57: Why are you removing the `Link`? This is unrelated to removing useRouter. And there's another instance of `Link` above in the code that wasn't removed.

marvinmarnold (changes requested):

app/components/ErrorBoundary/ErrorBoundary.tsx
yash431garg: I have the Link because we already have a button and instead of wrapping it with a Link I thought of just doing from the onClick method with using router.

danijel-radomiljac (dismissed):

app/components/ErrorBoundary/ErrorBoundary.tsx
marvinmarnold: Link doesn't use `useRouter` though. And like I said, line 43 still uses Link, so I don't get why change it just here.

More importantly, do you really want to link to `/failed`?

marvinmarnold (changes requested):

app/components/ErrorBoundary/ErrorBoundary.tsx
yash431garg: Sorry, I don't want to route to /failed it is a mistake I change to to home path.

marvinmarnold (approved):

PR #583, opened 2024-10-04
PROD-358: Enable captureConsole Integration Sentry

neel-ds:
- Description
Load Failed errors occur when Fetch API calls fail. It usually happens on Apple devices. Following the below documentation, I have added captureConsole integration to get additional context for such errors.
https://sentry.io/answers/load-failed-javascript/
- What are the steps to test that this code is working?
Sentry events will also show console logs of requests being made
- Screen shots or recordings for UI changes
NA

PR #597, opened 2024-10-13
PROD-384 Resolve load failed issue

neel-ds:
### Description

This issue was addressed in #583 and console errors sentry integration was added to assess such an issue. Load Failed errors occur when Fetch API calls fail. It usually happens on Apple devices. It’s not a disruptive error, but it gets triggered when the user opens the login modal and mixpanel returns 500. 

This PR updates the mixpanel route to always return 200 and disable console errors integration for developement environment to avoid unnecessary notifications in the Slack channel.

Mixpanel will always work technically and no more spam in sentry notifications.
Screenshot 2024-10-13 at 2 16 14 PM

NA


app/api/mixpanel/route.ts
```diff
@@ -48,19 +48,14 @@ export async function POST(request: NextRequest) {
...
 
-
     return NextResponse.json({ status: "Event tracked successfully" });
   } catch (error) {
-    console.log(error);
-    return NextResponse.json(
-      { error: "Internal Server Error" },
-      { status: 500 },
-    );
+    console.log("[Mixpanel]", error);
```
marvinmarnold, line 58: Can you make this error level log? Based `sentry.server.config.ts`, will a Sentry event be triggered if this is updated to an error level log?

marvinmarnold (changes requested):

danijel-radomiljac (approved):

app/api/mixpanel/route.ts
neel-ds: yes, sentry console integration will log this error

PR #602, opened 2024-10-15
PROD-395/welcome-back-screen-changes

danijel-radomiljac:


neel-ds (commented):
The chompy head icon resolution is poor. Can you use `unoptimized` prop and export the image with 2x size?

```ts
<Image
    src={url}
    alt={alt}
    width={width}
    height={height}
    unoptimized
/>
  ```

danijel-radomiljac:
> The chompy head icon resolution is poor. Can you use `unoptimized` prop and export the image with 2x size?
> 
> ```ts
> <Image
>     src={url}
>     alt={alt}
>     width={width}
>     height={height}
>     unoptimized
> />
> ```

I just added a bigger image. I think it is a not good idea to do unoptimization for images.

marvinmarnold (approved):

PR #608, opened 2024-10-16
PROD-403 Handle priority fee request failure

neel-ds:
### Description
It seems the reveal is occasionally failing due to the `getPriorityFeeEstimate` RPC method call failure, and this issue has regressed despite being resolved with the latest updates.

https://github.com/gator-labs/chomp/blob/5df2c285a899e339314163133b721bd042c395d6/app/queries/getPriorityFeeEstimate.ts#L14-L25

This PR includes handling of RPC call failures by retrying the method twice. If the maximum attempts fail, it will throw an error. Additionally, it logs console errors to help identify the cause of this intermittent behavior.
Better to monitor it and take necessary steps later
NA


app/utils/solana.ts
```diff
@@ -40,8 +40,21 @@ export const genBonkBurnTx = async (
...
+  // Verify the estimateFee is not null due to RPC request failure in some cases
+  if (estimateFee === null) {
+    for (let i = 0; i < 2; i++) {
+      estimateFee = await getRecentPrioritizationFees(tx);
+      if (estimateFee !== null) break;
+    }
+    
+    // Throw an error if the calls fail after maximum attempts
+    if (estimateFee === null) {
+      throw new Error("Failed to estimate priority fee after maximum attempts");
```
yash431garg, line 54: Instead of throwing an error we can add a fix priority fees incase estimate fee return null.

neel-ds: We can think of a reasonable number.

marvinmarnold (dismissed):

app/utils/solana.ts
```diff
@@ -40,8 +40,21 @@ export const genBonkBurnTx = async (
...
 
-  const estimateFee = await getRecentPrioritizationFees(tx);
+  let estimateFee = await getRecentPrioritizationFees(tx);
 
+  // Verify the estimateFee is not null due to RPC request failure in some cases
+  if (estimateFee === null) {
+    for (let i = 0; i < 2; i++) {
+      estimateFee = await getRecentPrioritizationFees(tx);
+      if (estimateFee !== null) break;
+    }
```
marvinmarnold, line 50: Ideally, should sleep here for 2s

neel-ds:
Added median priority fee if maximum attempts reach and delay of 2 secs between re-trials.

cc: @marvinmarnold @yash431garg

yash431garg:
> Added median priority fee if maximum attempts reach and delay of 2 secs between re-trials.
> 
> cc: @marvinmarnold @yash431garg

Looks fine to me how much time does it take incase when we use 5000 as estimatefee

neel-ds:
> > Added median priority fee if maximum attempts reach and delay of 2 secs between re-trials.
> > cc: @marvinmarnold @yash431garg
> 
> Looks fine to me how much time does it take incase when we use 5000 as estimatefee?

6-7 secs

yash431garg (dismissed):

marvinmarnold (approved):

yash431garg (approved):

PR #613, opened 2024-10-18
PROD-352: configured eslint, prettier and husky

yash431garg:
- Description
Integrated eslint for lint error includes tslint, import and hooks warning. Added prettier with formatting rules and husky for precommit check
No functional changes
N/A


.prettierignore
```diff
@@ -0,0 +1,10 @@
+node_modules/
+.next/
+public/
+out/
+coverage/
+build/
+.vercel/
+docs/
+prisma/migrations/
+
```
neel-ds, line 10: You can add more folders and files in ignore like:
- all files / folders starting with .
- yarn.lock
- postgres-data
- README files

.storybook/README.md
```diff
@@ -1,7 +1,7 @@
 ### Storybook
 
 When creating UI components developer should take care to run the command:
-```yarn run storybook```
+`yarn run storybook`
```
neel-ds, line 4: This change should be reverted once you follow up on `prettierignore` review

app/components/Icons/DeckIcon.tsx
```diff
@@ -1,7 +1,6 @@
 import { ICON_DEFAULT_WIDTH_HEIGHT, IconProps } from ".";
 
 export function DeckIcon({
-  fill = "#fff",
```
neel-ds, line 4: Can we keep this fill prop unless by default is #fff? Not sure if it was intentional

neel-ds (changes requested):
Kindly add more files and folders in `prettierignore` as mentioned below and revert the changes of same files or folders. I also found line breaks between imports in almost every files. What's the purpose behind it? Please remove that rule unless there is convention

yash431garg:
> Kindly add more files and folders in `prettierignore` as mentioned below and revert the changes of same files or folders. I also found line breaks between imports in almost every files. What's the purpose behind it? Please remove that rule unless there is convention

The reason behind line breaks in import was to separate packages which are external and internal components.

app/components/Icons/DeckIcon.tsx
yash431garg: It was not used here that why comes to property of unused vars.

.husky/pre-commit
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
marvinmarnold, line 1: Will all of these also be run when a PR is open?

yash431garg: No, It will run after we commits and will block the commit if there is some issue.

yash431garg: Sure, I will add this in the workflow.

marvinmarnold:
@neel-ds can you please re-review?

marvinmarnold (approved):

marvinmarnold (dismissed):

.prettierignore
neel-ds: To ignore README files, `*.md` use this instead of README*

neel-ds (dismissed):

marvinmarnold (approved):

neel-ds (approved):

marvinmarnold (approved):

PR #621, opened 2024-10-22
PROD-398 Add device ID for Identity management

neel-ds:
### Description
This PR implements server-side Identity Management to track events or UTM for pre-login and post-login users. It will allow us to track both kinds of users and connect the identities. Helpful in lead conversion and other metrics. `$device_id` will be a common attribute between pre-login and post-login.
Open various pages with UTM parameters and interact with them. Then, log in and interact with the reveal feature. After that, check the Mixpanel dashboard to ensure those events are being tracked correctly.
### Screenshots or recordings for UI changes
Pre-login
Screenshot 2024-10-22 at 4 48 37 PM
Post-login
Screenshot 2024-10-22 at 4 48 27 PM

### DIFF
--- a/app/api/mixpanel/route.ts
+++ b/app/api/mixpanel/route.ts
@@ -4,6 +4,8 @@ import { getCurrentUser } from "@/app/queries/user";
 import { kv } from "@/lib/kv";
 import Mixpanel from "mixpanel";
 import { ReadonlyHeaders } from "next/dist/server/web/spec-extension/adapters/headers";
+import { v4 as uuidv4 } from 'uuid';
+import { cookies } from "next/headers";
 
 const mixpanel = Mixpanel.init(process.env.MIX_PANEL_TOKEN!);
 
@@ -23,6 +25,49 @@ function getIPAddress(headersList: ReadonlyHeaders) {
   return headersList.get("x-real-ip") ?? FALLBACK_IP_ADDRESS;
 }
 
+async function handleUtmParams(userId: string, properties: any) {
+  const utmParams = {
+    utm_source: properties.$utm_source,
+    utm_medium: properties.$utm_medium,
+    utm_campaign: properties.$utm_campaign,
+    utm_term: properties.$utm_term,
+    utm_content: properties.$utm_content,
+  };
+
+  // Get stored UTM data from KV
+  const storedUtmData = await kv.get(`utm:${userId}`) as UtmData | null;
+  let initialUtm = storedUtmData?.initial_utm || {};
+  let lastUtm = storedUtmData?.last_utm || {};
+
+  // Check if any new UTM parameters are provided
+  const hasNewUtmParams = Object.values(utmParams).some(
+    (value) => value !== undefined,
+  );
+
+  const utmParamsWithoutUndefined = Object.fromEntries(
+    Object.entries(utmParams).filter(([_, value]) => value !== undefined)
+  );
+
+  // Update initial and last UTM data
+  if (!storedUtmData || hasNewUtmParams) {
+    initialUtm = Object.keys(initialUtm).length > 0 ? initialUtm : utmParamsWithoutUndefined;
+    lastUtm = hasNewUtmParams ? utmParamsWithoutUndefined : lastUtm;
+
+    // Remove undefined values
+    Object.keys(lastUtm).forEach(
+      (key) => lastUtm[key] === undefined && delete lastUtm[key],
+    );
+
+    // Store updated UTM data in KV
+    await kv.set(`utm:${userId}`, {
+      initial_utm: initialUtm,
+      last_utm: lastUtm,
+    });
+  }
+
+  return { initialUtm, lastUtm };
+}
+
 export async function POST(request: NextRequest) {
   const data = await request.json();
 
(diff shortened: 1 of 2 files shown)


app/api/mixpanel/route.ts
```diff
@@ -32,24 +77,51 @@ export async function POST(request: NextRequest) {
...
     } else {
-      const utmParams = {
-        utm_source: properties.$utm_source,
-        utm_medium: properties.$utm_medium,
-        utm_campaign: properties.$utm_campaign,
-        utm_term: properties.$utm_term,
-        utm_content: properties.$utm_content,
-      };
+      // Post-login user tracking
+      const { initialUtm, lastUtm } = await handleUtmParams(currentUser.id, properties);
```
marvinmarnold, line 124: User UTM should be used if present, otherwise use device UTM.

marvinmarnold (changes requested):

app/api/mixpanel/route.ts
neel-ds: Done

marvinmarnold (changes requested):
Can you add more test cases for all the scenarios we talked about.
- normal: logged out user visits, then confirms
- slow conversion: guest visits from Twitter, then later from Instagram and signs up
- logged in users: codify in a test what the behavior should be if user is logged in first time UTM is tracked, then signs out from the same device

Make clear assertions about the state of first and last utm for each case.

Let's get this into the next release.

neel-ds:
> Can you add more test cases for all the scenarios we talked about.
> 
> * normal: logged out user visits, then confirms
> * slow conversion: guest visits from Twitter, then later from Instagram and signs up
> * logged in users: codify in a test what the behavior should be if user is logged in first time UTM is tracked, then signs out from the same device
> 
> Make clear assertions about the state of first and last utm for each case.
> 
> Let's get this into the next release.

I added three more test cases specific to UTM tracking which asserts edge cases as mentioned above. Fixed some mistakes in route as well. Refactored codebase to meet standard practices as suggested.

app/api/mixpanel/route.ts
```diff
@@ -23,6 +25,57 @@ function getIPAddress(headersList: ReadonlyHeaders) {
...
+  const hasNewUtmParams = Object.values(utmParams).some(
+    (value) => value !== undefined,
+  );
+
+  const utmParamsWithoutUndefined = Object.fromEntries(
+    Object.entries(utmParams).filter(([_, value]) => value !== undefined)
+  );
+
+  // Update initial and last UTM data
+  if (hasNewUtmParams) {
```
yash431garg, line 57: Here you can directly check from (utmParamsWithoutUndefined) that if some value is defined and you don't need to do this extra step (hasNewUtmParams).

app/api/mixpanel/route.ts
```diff
@@ -32,77 +85,66 @@ export async function POST(request: NextRequest) {
...
+    let deviceId = cookies().get('device_id')?.value;
+    if (!deviceId) {
+      deviceId = uuidv4();
+    }
+
+    // Set the device ID cookie
+    cookies().set('device_id', deviceId, { 
+      httpOnly: true, 
+      secure: process.env.NODE_ENV === 'production',
+      maxAge: 60 * 60 * 24 * 365 // 1 YEAR
```
yash431garg, line 98: Not sure about the reason behind it. But Isn't 1year to long?

app/api/mixpanel/route.ts
```diff
@@ -23,6 +25,57 @@ function getIPAddress(headersList: ReadonlyHeaders) {
...
+  // Always prioritize device's initial UTM if it exists
+  let initialUtm = deviceUtmData?.initial_utm || userUtmData?.initial_utm || {};
+  let lastUtm = deviceUtmData?.last_utm || userUtmData?.last_utm || {};
+
+  // Check if any new UTM parameters are provided
+  const hasNewUtmParams = Object.values(utmParams).some(
+    (value) => value !== undefined,
+  );
+
+  const utmParamsWithoutUndefined = Object.fromEntries(
```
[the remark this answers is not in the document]
neel-ds: Thanks for the suggestion

app/api/mixpanel/route.ts
neel-ds: Good catch!

app/api/mixpanel/route.ts
neel-ds: Default behavior: If you don't set a maxAge or expires attribute, the cookie becomes a "session cookie". This means it will be deleted when the browser session ends (i.e., when the user closes their browser).

So ideally to have trace for longer period, it's good to have a year even if user revisits in months or so

yash431garg: Thanks for the clarification!

yash431garg (approved):

__tests__/api/mixpanel/route.test.ts
```diff
@@ -0,0 +1,291 @@
...
+        {
+          createdAt: new Date(),
+          updatedAt: new Date(),
+          address: "8Tj2Lx72kLttL7iCN5dsPQGTUzjQLpyctVcnV9oH8sPM",
+          userId: "a19f7611-5cd2-49b8-ad81-69dc42a2a5a3",
+        },
+      ],
+    };
+
+    await kv.set("utm:550e8400-e29b-41d4-a716-446655440000", {});
```
marvinmarnold, line 86: What is the purpose of these lines?

__tests__/api/mixpanel/route.test.ts
```diff
@@ -0,0 +1,291 @@
...
+    expect(responseData).toEqual({ status: "Event tracked successfully" });
+    const utmPostLogin = await kv.get(
+      "utm:a19f7611-5cd2-49b8-ad81-69dc42a2a5a3",
+    );
+    expect(utmPostLogin).toEqual({
+      initial_utm: { utm_source: "telegram", utm_medium: "community" },
+      last_utm: { utm_source: "twitter", utm_medium: "social" },
+    });
+  });
+});
```
marvinmarnold, line 291: Could you add a test case where:
1. user is logged out to start
2. then logs in (assert UTM carries over)
3. user logs in from a different device (assert UTM carries over)

marvinmarnold (changes requested):

__tests__/api/mixpanel/route.test.ts
neel-ds: The purpose of these lines is to clear the values for deviceId and userId before each test. If previous values remain, they are mistakenly treated as initial values in the subsequent tests.

__tests__/api/mixpanel/route.test.ts
neel-ds: Added!

PR #624, opened 2024-10-23
PROD-420/share-claim-all 

danijel-radomiljac:


app/a/[txHash]/page.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+  return {
+    openGraph: {
+      images,
+    },
+  };
+}
+
+export default async function Page() {
+  redirect(
```
Comment (marvinmarnold) [inline app/a/[txHash]/page.tsx:23]: This is correct according to the issue, but should be updated.

Some of these values will be different for links that are manually copied, vs ones that are shared via twitter.

I think utm_campaign and utm_content will be the same in both cases and can be appended here.

utm_source and utm_medium will need to be supplied on the link shown to users. @Yihwan what values do you want to use for the manual link copy case? 

@danijel-radomiljac one final thought is that it may be possible to infer Twitter shares via some other mechanism. For example, maybe we can look for the HTTP referrer. In that case, we could continue providing links without any UTM and add here like you are doing now.

app/api/og/share-claim-all/route.tsx
```diff
@@ -0,0 +1,133 @@
+/* eslint-disable @next/next/no-img-element */
+import { ChompyGuitarIcon } from '@/app/components/Icons/ChompyGuitarIcon';
+import prisma from '@/app/services/prisma';
+import { decryptString } from '@/app/utils/crypto-js';
+import { ImageResponse } from 'next/og';
+
+export async function GET(request: Request) {
+  const { searchParams } = new URL(request.url);
+  const encryptedHash = searchParams.get('txHash');
```
marvinmarnold, line 9: Lets change the param name to encryptedTxHash for clarity

lib/urls.ts
```diff
@@ -6,3 +8,10 @@ export const getDeckPath = (deckId: string | number) =>
 export const STACKS_PATH = "/stacks";
 export const LEADERBOARD_PATH = `${HOME_PATH}/leaderboard`;
 export const ANSWER_PATH = `${HOME_PATH}/answer`;
+export const getOgShareClaimAllPath = (txHash: string) =>
```
marvinmarnold, line 11: Change params to encryptedTxHash

lib/urls.ts
```diff
@@ -6,3 +8,10 @@ export const getDeckPath = (deckId: string | number) =>
...
+  `/api/og/share-claim-all?txHash=${txHash}`;
+export const getClaimAllShareUrl = (txHash: string) => {
+  const encryptedTxHash = encrypt(txHash);
```
marvinmarnold, line 14: These hashes will be long. How about we take the first X characters and encrypt just those?

app/constants/tracking.ts
```diff
@@ -1,76 +1,79 @@
...
+  QUESTION_ANSWERED_FAILED: 'QuestionAnsweredFailed',
+  DECK_COMPLETED: 'DeckCompletionSucceeded',
+  SHARE_EARN_URL: 'ShareTaskonURL',
+  HOME_STAT_CARD_DIALOG_OPENED: 'HomeStatCardDialogOpened',
+  HOME_STAT_CARD_DIALOG_CLOSED: 'HomeStatCardDialogClosed',
+  PAGE_VIEW: 'PageView',
+  WELCOME_BACK_ANSWER_BUTTON_CLICKED: 'WelcomeBackAnswerButtonClicked',
+  WELCOME_BACK_REVEAL_BUTTON_CLICKED: 'WelcomeBackRevealButtonClicked',
+  DECK_CLICKED: 'DeckClicked',
+  SHARE_DIALOG_LOADED: 'ShareDialogLoaded',
```
marvinmarnold, line 33: I think we should change all these to ShareAllDiaglogLoaded, etc. To disambiguate from the share single flow. Could also use the same event for both all and single, but supply a param to disambiguate.

marvinmarnold (changes requested):
Looks good. Very thorough to add the encryption, nice. 

You followed the issue requirements well, but I have some new thoughts.

app/a/[txHash]/page.tsx
Comment (danijel-radomiljac) [reply app/a/[txHash]/page.tsx:22]: Hmm I am not actually sure what I need to change

Comment (marvinmarnold) [reply app/a/[txHash]/page.tsx:22]: Yes that was rather wordy. 

1. Always append: utm_campaign=app_share&utm_content=claim_all
2. Check for [HTTP referrer](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer).
3. If referrer = twitter, then append utm_source=x&utm_medium=social
4. If referrer is something els, then append utm_source=REFERRER and utm_medium=manual_link

But I'm not sure about the values to use for step #4. @Yihwan may suggest other values.

app/actions/claim.ts
```diff
@@ -1,20 +1,20 @@
...
+import { sendBonk } from '../utils/bonk';
+import { ONE_MINUTE_IN_MILLISECONDS } from '../utils/dateUtils';
+import { ClaimError } from '../utils/error';
+import { acquireMutex } from '../utils/mutex';
+import { getBonkBalance, getSolBalance } from '../utils/solana';
+import { getJwtPayload } from './jwt';
 
 export async function claimQuestion(questionId: number) {
-  console.log("claim questions fired");
+  console.log('claim questions fired');
```
yash431garg, line 17: Is this console intentional?

yash431garg, line 17: Is this console intentional?

app/api/og/share-claim-all/route.tsx
```diff
@@ -0,0 +1,136 @@
...
+          <div tw="flex w-[10px]"></div>
+          <div tw="flex flex-col grow">
+            <ChompyGuitarIcon width={220} height={200} />
+          </div>
+        </div>
+      </div>
+    ),
+    {
+      width: 1200,
+      height: 630,
```
yash431garg, line 128: I think you need to change the dimensions here to maintain image quality. It is W: 622 & H: 418 in figma

app/utils/crypto-js.ts
yash431garg, line 1: Can you add comments in this file regarding the functions?

app/actions/claim.ts
danijel-radomiljac: yes

app/api/og/share-claim-all/route.tsx
danijel-radomiljac: og image needs to be 1200x630

app/utils/crypto-js.ts
danijel-radomiljac: it is just basic encryption and decryption

app/api/og/share-claim-all/route.tsx
marvinmarnold: Why does it have to be that size? I see that it is recommended: https://vercel.com/docs/functions/og-image-generation#technical-details

marvinmarnold (changes requested):
Just tested. Pretty smooth.

- Can we decrease the encrypted hash to 10 characters? How long is the unencrypted text in that case? Links are still long: https://chomp-jdteee2lr-gator-labs.vercel.app/a/VTJGc2RHVmtYMStzT3M0em9aUHA4ZXpvazR6WG5rd0FkLytYSkQ5OWd0Zz0
- Waiting on clarity on image size issue

danijel-radomiljac:
> Just tested. Pretty smooth.
> 
> * Can we decrease the encrypted hash to 10 characters? How long is the unencrypted text in that case? Links are still long: [chomp-jdteee2lr-gator-labs.vercel.app/a/VTJGc2RHVmtYMStzT3M0em9aUHA4ZXpvazR6WG5rd0FkLytYSkQ5OWd0Zz0](https://chomp-jdteee2lr-gator-labs.vercel.app/a/VTJGc2RHVmtYMStzT3M0em9aUHA4ZXpvazR6WG5rd0FkLytYSkQ5OWd0Zz0)
> * Waiting on clarity on image size issue

the shortes encryption result is 50chars.

Yea it is just the recommendation to keep 1200x630 dimension. Honestly it looks ok to me. 
Screenshot 2024-10-28 at 12 24 26

marvinmarnold (changes requested):
Blocked by:
- https://linear.app/gator/issue/PROD-420/share-claim-all#comment-6db3f238
- https://linear.app/gator/issue/PROD-420/share-claim-all#comment-e622798f

PR #629, opened 2024-10-28
PROD-423: added fee estimation for bonk claim transaction

yash431garg:
- Description
Added priority fees, compute unit calculate logic and retry transaction in bonk claim
Greater chances of successful claim after reveal
N/A


app/utils/bonk.ts
```diff
@@ -38,35 +41,94 @@ export const sendBonk = async (
...
+
+      txtSig = await CONNECTION.sendRawTransaction(
+        updatedVersionedTransaction.serialize(),
+        {
+          skipPreflight: skipPreflightChecks,
+          maxRetries: 0,
+        },
+      );
+
+      return await pollTransactionConfirmation(txtSig);
```
marvinmarnold, line 123: I think this could lead to us sending multiple transactions. In a case where the tx succeeds, but the confirmations fails for some reason.

In that case, a special Sentry error could be triggered with the tx signature. User should see an error message that support is looking into it. We can manually send funds if it looks like tx really failed.

app/utils/bonk.ts
```diff
@@ -38,35 +41,94 @@ export const sendBonk = async (
...
+  const blockhashValidityThreshold = 4;
+
+  let retryCount: number = 0;
+  let txtSig: string;
+
+  const maxRetries: number = 6;
+  const skipPreflightChecks: boolean = true;
+
+  // Send the transaction with configurable retries and preflight checks
+  while (retryCount <= maxRetries) {
```
marvinmarnold, line 105: Consider using a 3rd party library or abstracting this code into a utility method if it comes up a lot.

eg https://www.npmjs.com/package/p-retry

marvinmarnold (changes requested):
I was surprised to see this approach differ so much from the existing reveal code. Was this pattern inspired by some source? gpt?

If you think this is bettter than the reveal code, should we update that code too?

app/utils/bonk.ts
```diff
@@ -38,35 +41,94 @@ export const sendBonk = async (
...
+    microLamports: estimateFee.priorityFeeEstimate || 0,
   });
 
-  await CONNECTION.confirmTransaction(
-    {
-      signature,
-      ...blockhash,
-    },
-    "confirmed",
+  instructions.unshift(computeBudgetIx);
```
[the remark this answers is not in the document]
yash431garg: This will update the dynamic generated instructions for the compute unit

app/utils/bonk.ts
```diff
@@ -38,35 +41,94 @@ export const sendBonk = async (
...
+
+  // Get the optimal compute units
+  const unitsConsumed = await getComputeUnits(
```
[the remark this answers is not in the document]
yash431garg: It was suggested on the official docs to calculate compute unit dynamically as well.

app/utils/bonk.ts
yash431garg: I will look into it.

yash431garg:
> I was surprised to see this approach differ so much from the existing reveal code. Was this pattern inspired by some source? gpt?
> 
> If you think this is bettter than the reveal code, should we update that code too?

It was inspired by official helius, solana and quicknode docs.

app/utils/bonk.ts
yash431garg: Thansks, for the suggestion!

app/utils/bonk.ts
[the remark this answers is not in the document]
marvinmarnold: I still think this should be removed. Use hardcoded value instead.

neel-ds (dismissed):

app/utils/bonk.ts
```diff
@@ -12,6 +12,7 @@ import {
   VersionedTransaction,
```
marvinmarnold, line 12: Can we rename this file while at it? Maybe `claim.ts`?

app/utils/bonk.ts
```diff
@@ -38,28 +39,58 @@ export const sendBonk = async (
...
-  const transaction = new VersionedTransaction(message);
+  const versionedTransaction = new VersionedTransaction(v0message);
+
+  versionedTransaction.sign([fromWallet]);
+
+  const estimateFee = await getRecentPrioritizationFees(versionedTransaction);
+
+  // Add the compute unit price instruction with the estimated fee
+  const computeBudgetIx = ComputeBudgetProgram.setComputeUnitPrice({
+    microLamports: estimateFee?.result?.priorityFeeLevels?.high || 5000,
```
marvinmarnold, line 60: Where are you getting 5,000 from here? Looking at some recent burns (where we do dynamically calculate price), prices range from 50,000 - 200,000.

https://solscan.io/tx/2gMj99sRHNh1n9cwxFu1nN2MKnHT1vZep9NLmW4Vcm8UbEVTtna2AE4UNs5FGDLSQbBwaKn7obEbjvJQz4eqxHpj

https://solscan.io/tx/2Tu1j4CCgt3zSMQEcELLtun5vQS6uUnmwRJccwDU4U5H4HVqYHsiVEG1tHF8UDshMJGpeRa3bcB5gzywCmBnu9Yv

https://solscan.io/tx/3FfzRhSeWHXgzPFkCC3BeWu2Qxnw7t3evMrfcw1Ci546WmWJDnUeDvYXYjJbWkyn1Ga8KCpdD4h7nKgqnD3CAKfE

app/utils/bonk.ts
```diff
@@ -38,28 +39,58 @@ export const sendBonk = async (
...
+  });
+
+  // update the instructions with compute budget instruction.
+  instructions.unshift(computeBudgetIx);
 
-  transaction.sign([fromWallet]);
+  const computeUnitFix = 5000;
```
marvinmarnold, line 66: Should be 4794.

app/utils/bonk.ts
```diff
@@ -38,28 +39,58 @@ export const sendBonk = async (
...
 
-  const signature = await CONNECTION.sendTransaction(transaction, {
-    maxRetries: 10,
+  const computeUnitsIx = ComputeBudgetProgram.setComputeUnitLimit({
+    units: computeUnitFix * 1.1,
```
marvinmarnold, line 69: Maybe we can get rid of the buffer (1.1). The units never seems to change.

marvinmarnold (changes requested):

app/actions/leaderboard.ts
```diff
@@ -164,11 +164,11 @@ export const getLeaderboard = async ({
 const getNumberOfChompedQuestions = async (
   dateFilter:
     | {
-      createdAt: {
-        gte: Date;
-        lte: Date;
-      };
-    }
+        createdAt: {
```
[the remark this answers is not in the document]
yash431garg: Yea, this is an old pr that's why we still have that formatting related changes here.

app/queries/getComputeUnitEstimate.ts
```diff
@@ -0,0 +1,46 @@
...
+import { CONNECTION } from "../utils/solana";
+
+/**
+ * Simulate a transaction to get the total compute units consumed
+ * @param {TransactionInstruction[]} instructions - The transaction instructions
+ * @param {PublicKey} payer - The public key of the payer
+ * @param {AddressLookupTableAccount[]} lookupTables - The address lookup tables
+ * @returns {Promise<number | null>} - The compute units consumed, or null if unsuccessful
+ */
+export const getComputeUnits = async (
```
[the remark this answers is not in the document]
yash431garg: Sure, I thought that it can be used in future but will remove it for now.

app/utils/bonk.ts
yash431garg: Sure,

app/utils/bonk.ts
yash431garg: Fair, I also update the estimate to 50k if estimateFee query fails

app/utils/bonk.ts
yash431garg: Adding compute unit in exact number cause the transaction to fail that's why a buffer is added.

app/utils/bonk.ts
yash431garg: Done

app/utils/claim.ts
```diff
@@ -0,0 +1,103 @@
...
+  updatedVersionedTransaction.sign([fromWallet]);
+
+  const signature = await CONNECTION.sendTransaction(
+    updatedVersionedTransaction,
+    {
+      maxRetries: 10,
+    },
+  );
+
+  await CONNECTION.confirmTransaction(
```
marvinmarnold, line 94: What happens if tx goes through but can't be confirmed for some reason? What will the user see?

app/utils/claim.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  // Add the compute unit price instruction with the estimated fee
+  const computeBudgetIx = ComputeBudgetProgram.setComputeUnitPrice({
+    microLamports: estimateFee?.result?.priorityFeeLevels?.high || 50000,
+  });
+
+  // update the instructions with compute budget instruction.
+  instructions.unshift(computeBudgetIx);
+
+  const computeUnitFix = 4794;
```
marvinmarnold, line 66: Add a comment about where this number comes from

app/utils/claim.ts
```diff
@@ -0,0 +1,103 @@
...
+
+  const versionedTransaction = new VersionedTransaction(v0message);
+
+  versionedTransaction.sign([fromWallet]);
+
+  const estimateFee = await getRecentPrioritizationFees(versionedTransaction);
+
+  // Add the compute unit price instruction with the estimated fee
+  const computeBudgetIx = ComputeBudgetProgram.setComputeUnitPrice({
+    microLamports: estimateFee?.result?.priorityFeeLevels?.high || 50000,
```
marvinmarnold, line 60: Add comment about where 50000 comes from

marvinmarnold (dismissed):

app/utils/claim.ts
marvinmarnold: Make this a variable and reuse in burn

app/utils/claim.ts
```diff
@@ -0,0 +1,103 @@
...
+  });
+
+  // update the instructions with compute budget instruction.
+  instructions.unshift(computeBudgetIx);
+
+  const computeUnitFix = 4794;
+
+  const computeUnitsIx = ComputeBudgetProgram.setComputeUnitLimit({
+    units: computeUnitFix * 1.1,
```
marvinmarnold, line 69: Can you leave a comment that this fails unless there is buffer

app/utils/claim.ts
yash431garg: This https://github.com/gator-labs/chomp/blob/814ff3a81e597873660992a70dfc9588d01a036a/app/components/History/TotalRewardsClaimAll/TotalRewardsClaimAll.tsx#L56

marvinmarnold: image does not load

app/utils/solana.ts
```diff
@@ -55,17 +56,18 @@ export const genBonkBurnTx = async (
...
-            high: 5000,
+            high: HIGH_PRIORITY_FEE,
           },
         },
       };
     }
   }
 
-  const computeUnitFix = 5000;
+  const computeUnitFix = 4794;
```
marvinmarnold, line 66:
```suggestion
  const computeUnitFix = 4960;
```

app/utils/claim.ts
yash431garg: User will see Transaction Failed! Please try again. If this issue keep happening please, let us know on telegram.

PR #630, opened 2024-10-28
Prod 422: error recovery failed claim

yash431garg:
- Description
Previously if claim fails user isn't able to claim again but his bonk is already burned after this change user will be able do claim again incase it fails.
Error through if sendBonk failed and not user should be able to claim again if it fails
Screenshot 2024-10-28 at 5 43 07 PM

### DIFF
--- a/app/providers/ToastProvider.tsx
+++ b/app/providers/ToastProvider.tsx
@@ -8,6 +8,7 @@ import { InfoIcon } from "../components/Icons/ToastIcons/InfoIcon";
 import { RemoveIcon } from "../components/Icons/ToastIcons/RemoveIcon";
 import { SpinnerIcon } from "../components/Icons/ToastIcons/SpinnerIcon";
 import { SuccessIcon } from "../components/Icons/ToastIcons/SuccessIcon";
+import Link from "next/link";
 
 type ToastContextType = {
   successToast: (message: string, description?: string) => void;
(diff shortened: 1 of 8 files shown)


danijel-radomiljac (dismissed):

marvinmarnold (changes requested):
Ah I forgot to submit this comment.

app/providers/ToastProvider.tsx
```diff
@@ -77,8 +79,35 @@ const successToastLayout = (message: string, description?: string) =>
 const infoToastLayout = (message: string, description?: string) =>
   toastLayout(InfoIcon, message, description);
 
-const errorToastLayout = (message: string, description?: string) =>
-  toastLayout(ErrorIcon, message, description);
+const errorToastLayout = (message: string, description?: string) => {
+  const hasTelegramLink = message.includes(TELEGRAM_LINK_MARKER);
+  const parts = message.split(TELEGRAM_LINK_MARKER);
```
marvinmarnold, line 84: I still don't love this code. At the minimum, we also need comments here and in the error definition to explain the format.

After looking at the code more closely, we could update `errorToastLayout` to accept an additional optional param.

```
interface IToastExtraConfig= {}

interface IToastTelegramConfig extends IToastExtraConfig {
   type: 'telegram_link',
   text: string
}

errorToastLayout = (message: string, description?: string, extraConfig?: IToastExtraConfig)
```

Just one possible implementation..

yash431garg: Thanks, I spent a little more time on it and updated the ToastProvider so that it can now accept ReactNode and now we don't need to keep that implementation.

PR #633, opened 2024-10-29
PROD-408 Routes for notification and bot user

neel-ds:
- Description
The PR adds two routes for notification handling and bot user handling. It includes schema changes required to maintain subscriber property. 

Test the bot with /resubscribe and /unsubscribe commands

Screenshot 2024-10-30 at 2 36 53 AM


prisma/schema.prisma
```diff
@@ -26,6 +27,7 @@ model User {
   fungibleAssetTransactionLog FungibleAssetTransactionLog[]
   revealNfts                  RevealNft[]
   tutorialCompletedAt         DateTime?
+  isSubscriber                Boolean                       @default(false)
```
marvinmarnold, line 30: This variable name is not specific enough. Can you rename to `isBotSubscriber` or similar.

app/api/notification/route.ts
```diff
@@ -0,0 +1,71 @@
+/*
+  THIS ROUTE CONTAINS TWO METHODS GET AND POST: 
+  ➤ GET METHOD: FETCHES SUBSCRIBED USERS
```
marvinmarnold, line 3: Let's split this up into two files. Paths should be more descriptive, eg `app/api/users/getSubscribedUsers/route.ts`

app/api/notification/route.ts
```diff
@@ -0,0 +1,71 @@
...
+  ➤ POST METHOD: SETS A USER AS SUBSCRIBER
+*/
+import prisma from "@/app/services/prisma";
+import { headers } from "next/headers";
+
+export async function GET() {
```
marvinmarnold, line 9: All API methods should have tests

app/api/notification/route.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    return Response.json({ users });
+  } catch {
+    return new Response(`Error getting subscribed users`, {
+      status: 500,
+    });
+  }
+}
+
+export async function POST(request: Request) {
```
marvinmarnold, line 38: Unnecessary? Can't the user#GET route below be used for this?

marvinmarnold (changes requested):

__tests__/api/user/createUserByTelegramId/route.test.ts
```diff
@@ -0,0 +1,166 @@
+import { POST } from "@/app/api/user/createUserByTelegramId/route";
+import prisma from "@/app/services/prisma";
+import * as nodeCrypto from "crypto";
+import * as jwt from "jsonwebtoken";
+
+// Constants
+const VALID_API_KEY = process.env.BOT_API_KEY;
+const BOT_TOKEN = process.env.BOT_TOKEN ?? "";
+const BASE_URL = process.env.VERCEL_PROJECT_PRODUCTION_URL
```
marvinmarnold, line 9: You shouldn't need to define a remote endpoint. And should definitely not be using a prod endpoint. See: https://github.com/gator-labs/chomp/blob/7c80d01eebe9610b39ddd0f73f9e6f12b56d8861/__tests__/api/mixpanel/route.test.ts#L92

__tests__/api/user/createUserByTelegramId/route.test.ts
```diff
@@ -0,0 +1,166 @@
...
+      expect(data.profile.wallets).toEqual([]);
+
+      // Store ID for cleanup
+      createdUserId = data.profile.id;
+
+      // Verify user exists in database
+      const dbUser = await prisma.user.findUnique({
+        where: { id: createdUserId },
+      });
+      expect(dbUser).toBeDefined();
```
marvinmarnold, line 78: Would be nice to make some additional assertions about the fields that we expect in dbUser.

__tests__/api/user/createUserByTelegramId/route.test.ts
```diff
@@ -0,0 +1,166 @@
...
+
+      // Assert
+      expect(response.status).toBe(500);
+      expect(errorMessage).toBe("Error creating user");
+    });
+  });
+});
+
+// Helper function to encode the Telegram payload
+export async function encodeTelegramPayload(TELEGRAM_USER_PAYLOAD: any) {
```
marvinmarnold, line 127: Would be nice to put this in a test utility file

__tests__/api/user/getUserByTelegramId/route.test.ts
```diff
@@ -0,0 +1,164 @@
...
+
+describe("GET /api/user/getUserByTelegramId", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+    mockApiKey = VALID_API_KEY;
+  });
+
+  describe("when all parameters are valid", () => {
+    it("should return user profile with 200 status", async () => {
+      // Arrange
```
marvinmarnold, line 58: Nice, good to see the test split up into the three sections, for clarity.

__tests__/api/user/setUserSubscription/route.test.ts
```diff
@@ -0,0 +1,172 @@
...
+const VALID_API_KEY = process.env.BOT_API_KEY;
+const TELEGRAM_USER_PAYLOAD = {
+  id: "705689",
+  first_name: "John",
+  last_name: "",
+  username: "john_doe",
+  photo_url: "",
+  auth_date: String(Math.floor(new Date().getTime())),
+};
+const BASE_URL = process.env.VERCEL_PROJECT_PRODUCTION_URL
```
marvinmarnold, line 17: Same issue that this shouldn't be necessary.

app/api/user/createUserByTelegramId/route.ts
```diff
@@ -0,0 +1,90 @@
...
+  const data = await request.json();
+  const { telegramAuthToken } = data;
+
+  if (!telegramAuthToken) {
+    return Response.json("Telegram payload is required", {
+      status: 400,
+    });
+  }
+
+  const randomUUID = uuidv4();
```
marvinmarnold, line 51: This doesn't seem right to me. With our new direction, even TG users will be signed in before they use the app. We should never need to create a random UUID. It should always be available from Dynamic.

marvinmarnold (changes requested):

__tests__/api/user/createUserByTelegramId/route.test.ts
neel-ds: Good idea!

__tests__/api/user/createUserByTelegramId/route.test.ts
neel-ds: Added telegramId assertion

app/api/user/createUserByTelegramId/route.ts
neel-ds: I went through the Dynamic APIs and am unsure if we can create a user based on TG credentials.

__tests__/api/user/getUserByTelegramId/route.test.ts
```diff
@@ -0,0 +1,164 @@
+import verifyTelegramAuthToken from "@/app/actions/bot";
+import { GET } from "@/app/api/user/getUserByTelegramId/route";
+import prisma from "@/app/services/prisma";
+
+// Constants
+const VALID_API_KEY = process.env.BOT_API_KEY;
+const BASE_URL = process.env.VERCEL_PROJECT_PRODUCTION_URL
```
marvinmarnold, line 7: Why is this necessary?

prisma/migrations/20241029134128_add_tg_bot_fields/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- AlterTable
+ALTER TABLE "User" ADD COLUMN     "isBotSubscriber" BOOLEAN NOT NULL DEFAULT false,
+ADD COLUMN     "telegramId" BIGINT;
```
marvinmarnold, line 3: Should be distinct

app/api/user/createUserByTelegramId/route.ts
marvinmarnold: @neel-ds will update auto login code to always update randomly generated user ID with a Dynamic ID.

app/actions/jwt.ts
```diff
@@ -47,21 +47,69 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+      });
+    } else {
+      // If no existing user was found or the IDs match, create/update the user record
+      user = await prisma.user.upsert({
+        where: {
+          id: payload.sub,
+        },
+        create: {
+          id: payload.sub,
+          profileSrc: getRandomAvatarPath(),
```
marvinmarnold, line 86: This will result in us updating the profile image all the time.

app/actions/jwt.ts
```diff
@@ -47,21 +47,69 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
-    },
-    update: {},
-    include: {
-      wallets: true,
-      emails: true,
-    },
-  });
+  let user;
+
+  if (telegramUsername) {
```
marvinmarnold, line 52: I'm not sure this will keep working if a user changes their username

marvinmarnold (changes requested):

app/actions/jwt.ts
neel-ds: This is the existing implementation which I added in else statement. I think `upsert` didn't change the profile before as far as I know

neel-ds:
I updated the logic for storing users based on Telegram ID instead of Telegram username.

app/actions/jwt.ts
```diff
@@ -47,21 +51,71 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+    }
+  } else {
+    user = await prisma.user.upsert({
+      where: {
+        id: payload.sub,
+      },
+      create: {
+        id: payload.sub,
+        profileSrc: getRandomAvatarPath(),
+        telegramUsername: telegramUsername,
```
marvinmarnold, line 110: This will always be null here. Might as well remove, for clarity that nothing is being set.

marvinmarnold (changes requested):

app/actions/jwt.ts
```diff
@@ -47,21 +51,82 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+            id: payload.sub,
+            telegramUsername: telegramUsername,
+          },
+          include: {
+            wallets: true,
+            emails: true,
+          },
+        });
+      } else {
+        // If the user's Dynamic ID matches the JWT payload, then update their Telegram username incase user may have changed it
```
marvinmarnold, line 85: Why is it even necessary to update here? I would think that's only necessary if the username has changed.

neel-ds:
I optimized the conditional statements to handle the user data store (create or update). No additional calls will be made in any conditions if not necessary.

app/actions/jwt.ts
```diff
@@ -47,21 +51,75 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+        include: {
+          wallets: true,
+          emails: true,
+        },
+      });
+    }
+
+    // Update existing user if needed
+    const isTelegramUserDataChanged =
+      existingTelegramUser.id !== payload.sub ||
```
marvinmarnold, line 95: In what case would the tgId change? I don't think tgId should ever change.

marvinmarnold (commented):
A ew more small questions.

app/api/user/setUserSubscription/route.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const data = await request.json();
+  const { telegramAuthToken, isBotSubscriber } = data;
+
+  try {
+    const telegramUserData = (await verifyTelegramAuthToken(
+      telegramAuthToken,
+    )) as TelegramAuthDataProps;
+
+    const user = await prisma.user.findFirst({
```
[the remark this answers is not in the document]
neel-ds: Good catch

app/api/user/setUserSubscription/route.ts
```diff
@@ -0,0 +1,60 @@
...
+      where: { telegramId: telegramUserData.id },
+      select: {
+        id: true,
+      },
+    });
+
+    await prisma.user.update({
+      where: {
+        telegramUsername: telegramUserData.username,
```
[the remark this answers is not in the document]
neel-ds: Done

app/actions/jwt.ts
neel-ds: Telegram Id will always be the same.

app/actions/jwt.ts
```diff
@@ -47,21 +51,75 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+        update: {},
+        include: {
+          wallets: true,
+          emails: true,
+        },
+      });
+    }
+
+    // Check for existing Telegram user
+    const existingTelegramUser = await prisma.user.findFirst({
```
yash431garg, line 72: I think you can use upsert here as well and don't need to use separately insert and update.

app/actions/jwt.ts
```diff
@@ -47,21 +51,75 @@ export const setJwt = async (token: string, nextPath?: string | null) => {
...
+        },
+        include: {
+          wallets: true,
+          emails: true,
+        },
+      });
+    }
+
+    // If no updates needed, fetch current user
+    return prisma.user.findUnique({
```
yash431garg, line 113: Is this necessary?

yash431garg (commented):
Can you check these comments? I think it will be fine after that only getUserByTelegram test will need some updates.

yash431garg (approved):

marvinmarnold (approved):

PR #644, opened 2024-11-06
fix: check wallet object is defined before use.

samott:
Added some null/undefined checks - I observed come crashes locally when a wallet is not installed.

### DIFF
--- a/app/application/layout.tsx
+++ b/app/application/layout.tsx
@@ -21,7 +21,7 @@ export default async function Layout({ children }: PageLayoutProps) {
     getTransactionHistory(),
   ]);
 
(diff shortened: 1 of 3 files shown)


marvinmarnold (approved):
Thanks for proactively adding this.

marvinmarnold:
@danijel-radomiljac @AndrejRadomiljac @yash431garg @neel-ds please review

yash431garg:
> @danijel-radomiljac @AndrejRadomiljac @yash431garg @neel-ds please review

Make sense to me if in some case the wallet or address object is undefined it can cause error.

yash431garg (approved):

PR #645, opened 2024-11-06
PROD-450/allow-questions-to-be-revealed-with-only-3-answers 

danijel-radomiljac:
@marvinmarnold We can update MINIMAL_ANSWERS_PER_QUESTION to 5. Also I needed to divide questionAnswers.length with questionOptions.length in getTotalNumberOfDeckQuestions function. Because if we have multi choice question, and that question has 4 answers by one user. It is actually only one answer, because we are storing answer for each option of a question in a db.

### DIFF
--- a/app/actions/leaderboard.ts
+++ b/app/actions/leaderboard.ts
@@ -44,7 +44,7 @@ export const getPreviousUserRank = async (
     return;
 
   const currentUser = await getCurrentUser();
-  
+
   const dateRange = getDateRange(variant, true);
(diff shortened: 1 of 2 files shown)


marvinmarnold:
This looks good. I'll wait to merge until this is fixed: https://linear.app/gator/issue/PROD-446/prevent-rewards-calculation-error-when-few-results#comment-5707af47

Rewards calculation will fail if there are not enough 2nd order responses for the same option that the user responded to.

PR #647, opened 2024-11-07
PROD-453/share-all-fixes

danijel-radomiljac:


marvinmarnold (commented):
I just noticed that the image rendering logic is repeated in ClaimPreviewImage and share-claim-all/route.tsx. Why is that? Why can't route.tsx be used to render the preview image also?

PR #649, opened 2024-11-08
PROD-402: add db constraint on fungibleAssetTransactionLog.

yash431garg:
- Description
Previously it was possible to invoke the answer question endpoint and get as many point as user want to prevent that now we have db constraint to prevent duplicate entry for the same userId and questionId and for deckId and questionId
No duplicates answer point for the same deck or question id.
N/A

### DIFF
--- a/prisma/migrations/20241030104838_add_telegram_username/migration.sql
+++ b/prisma/migrations/20241030104838_add_telegram_username/migration.sql
@@ -1,2 +1,2 @@
 -- AlterTable
-ALTER TABLE "User" ADD COLUMN     "telegramUsername" TEXT UNIQUE; 
+ALTER TABLE "User" ADD COLUMN "telegramUsername" TEXT UNIQUE; 
--- a/prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
+++ b/prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
@@ -6,7 +6,7 @@
 
 */
 -- CreateIndex
-CREATE UNIQUE INDEX "FungibleAssetTransactionLog_userId_questionId_key" ON "FungibleAssetTransactionLog"("userId", "questionId");
+CREATE UNIQUE INDEX "FungibleAssetTransactionLog_type_userId_questionId_key" ON "FungibleAssetTransactionLog"("type", "userId", "questionId");
 
 -- CreateIndex
 CREATE UNIQUE INDEX "FungibleAssetTransactionLog_deckId_userId_key" ON "FungibleAssetTransactionLog"("deckId", "userId");
--- a/prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
+++ b/prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
@@ -1,12 +1,12 @@
 /*
   Warnings:
 
-  - A unique constraint covering the columns `[userId,questionId]` on the table `FungibleAssetTransactionLog` will be added. If there are existing duplicate values, this will fail.
-  - A unique constraint covering the columns `[deckId,userId]` on the table `FungibleAssetTransactionLog` will be added. If there are existing duplicate values, this will fail.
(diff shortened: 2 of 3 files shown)


danijel-radomiljac (approved):

prisma/migrations/20241030104838_add_telegram_username/migration.sql
```diff
@@ -1,2 +1,2 @@
 -- AlterTable
-ALTER TABLE "User" ADD COLUMN     "telegramUsername" TEXT UNIQUE; 
+ALTER TABLE "User" ADD COLUMN "telegramUsername" TEXT UNIQUE;
```
marvinmarnold, line 2: This looks like it was accidentally included.

prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
```diff
@@ -0,0 +1,12 @@
...
+
+  - A unique constraint covering the columns `[userId,questionId]` on the table `FungibleAssetTransactionLog` will be added. If there are existing duplicate values, this will fail.
+  - A unique constraint covering the columns `[deckId,userId]` on the table `FungibleAssetTransactionLog` will be added. If there are existing duplicate values, this will fail.
+
+*/
+-- CreateIndex
+CREATE UNIQUE INDEX "FungibleAssetTransactionLog_type_userId_questionId_key" ON "FungibleAssetTransactionLog"("type", "userId", "questionId");
+
+-- CreateIndex
+CREATE UNIQUE INDEX "FungibleAssetTransactionLog_deckId_userId_key" ON "FungibleAssetTransactionLog"("deckId", "userId");
```
marvinmarnold, line 12: Can you load the prod db to your local machine and confirm that this migration doesn't fail?

marvinmarnold (changes requested):

prisma/migrations/20241030104838_add_telegram_username/migration.sql
yash431garg: I thought there is some extra space that's why I removed that.

prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
yash431garg: It failed and I investigated little more and find that someone has done the exploit.

prisma/migrations/20241030104838_add_telegram_username/migration.sql
marvinmarnold: Ah i see OK. It's fine.

prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
marvinmarnold: As next steps, I'm asking the original reporter to confirm his wallet address. In the meantime, @yash431garg will look into a script for fixing these duplicates so that we can remove them and add the constraint.

prisma/migrations/20241108065629_add_unique_deckid_userid/migration.sql
```diff
@@ -0,0 +1,45 @@
...
+    SELECT oldest_ctid
+    FROM CTE_KeepOldest
+);
+
+-- Create unique index for `type`, `userId`, and `questionId`
+CREATE UNIQUE INDEX "FungibleAssetTransactionLog_type_userId_questionId_key" 
+ON "FungibleAssetTransactionLog"("type", "userId", "questionId");
+
+-- Create unique index for `type`, `deckId`, and `userId`
+CREATE UNIQUE INDEX "FungibleAssetTransactionLog_deckId_userId_key"
```
samott, line 44: shouldn't this be `FungibleAssetTransactionLog_type_deckId_userId_key`?

yash431garg: Yes, thanks for pointing out. I have updated.

samott (approved):

marvinmarnold (approved):

PR #650, opened 2024-11-08
PROD-448 Debounce Second Order Selection Tracking

neel-ds:
- Description
It adds debounce to avoid too many requests at same time if user drags the slider continuously. It will delay the tracking call by 0.5 secs after percentage value changes.

Answer second order and drag slider continuously. Verify the events recorded in mixpanel dashboard. It shouldn't hit too many requests at same time.

### DIFF
--- a/app/components/Deck/Deck.tsx
+++ b/app/components/Deck/Deck.tsx
@@ -219,6 +219,13 @@ export function Deck({
       }
 
       if (currentQuestionStep === QuestionStep.PickPercentage) {
+        trackQuestionAnswer(
+          question,
+          "SECOND_ORDER",
+          deckId,
+          deckVariant,
+          optionPercentage,
+        );
         setDeckResponse((prev) => {
           const newResponses = [...prev];
           const response = newResponses.pop();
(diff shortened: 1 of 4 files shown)


marvinmarnold (commented):
Can you convince me that debounce is correct? It may still track values if the user is holding down on the mouse, but hasn't selected. We don't care about intermediary slider values until the user does click up.

neel-ds:
> Can you convince me that debounce is correct? It may still track values if the user is holding down on the mouse, but hasn't selected. We don't care about intermediary slider values until the user does click up.

Answered in Linear

app/components/Deck/Deck.tsx
```diff
@@ -219,6 +219,13 @@ export function Deck({
       }
 
       if (currentQuestionStep === QuestionStep.PickPercentage) {
+        trackQuestionAnswer(
```
marvinmarnold, line 222: Why does the tracking also need to happen here? I'm not clear when this gets invoked.

neel-ds: This will be invoked when a user clicks on Chomp button after selecting the second order.

marvinmarnold: Won't this lead to double tracking? It will have already been tracked once on click up.

neel-ds: It won't do double tracking as I removed the tracking when users drag the slider.

PR #656, opened 2024-11-12
Fix/param validation

samott:
- Fixes issue with DB queries.


marvinmarnold:
Replaced by https://github.com/gator-labs/chomp/pull/657

PR #662, opened 2024-11-13
fix: remove tracking for second order mc

neel-ds:
- Description
Removes the tracking of SECOND_ORDER for multi-choice questions. 

NA

NA


marvinmarnold:
@neel-ds can you confirm that 2nd order tracking will still happen on click-up for multiple choice?

neel-ds:
> @neel-ds can you confirm that 2nd order tracking will still happen on click-up for multiple choice?

Sure

neel-ds:
> There are also conflicts here now fyi

I have resolved it. I confirm that it tracks the second order only once when the user clicks on chomp button. Verifiable from mixpanel staging events

PR #663, opened 2024-11-13
PROD-483/more-share-all-fixes

danijel-radomiljac:


danijel-radomiljac:
@marvinmarnold I replaced an image tag with ClaimPreviewImage, but og image takes some time to load. I added all optimization techinques for image, but it still takes some time to load. So maybe we can add a placeholder?

marvinmarnold:
> @marvinmarnold I replaced ClaimPreviewImage with image tag, but og image takes some time to load. I added all optimization techinques for image, but it still takes some time to load. So maybe we can add a placeholder?

Sure. Is there a loading icon we can use? We don't have a placeholder now I don't think, but we can get one in the future.

PR #664, opened 2024-11-13
fix: load timer after hydration completes

yash431garg:
- Description
Timer wasn't moving after the previous fixes so this pr make sure that we only show timer after client hydration is complete
N/A


samott (approved):
Makes sense

PR #667, opened 2024-11-13
PROD-441: Filter for questions ready to reveal in "Reveal" tab

samott:
- Filter for questions ready to reveal in "Reveal" tab.
- Add some tests for the history query.


samott:
The tests for this are failing but they work locally and I can't see where the problem lies - could use another set of eyes.

danijel-radomiljac (approved):

samott:
Think I know what the issue is... testing my hypothesis now

marvinmarnold:
@samott the issue is not with your code. it's that the pre-existing test code in deck-test assumed that there are no unanswered questions. i created unanswered questions while debugging. i can think of a few options:

1. manually delete existing questions
2. programmatically delete existing questions in test code
3. in deck-test.ts, create answers for all unanswered questions as part of `beforeAll`

My preference is #3 because it is a permanent fix and less likely to interfere with other tests in the future. Can you do that please?

marvinmarnold (approved):
Way to power through this

PR #673, opened 2024-11-19
Prod 476/myster box without token transfer

yash431garg:
- Description
This PR will add the breadboard for mystery box spike for staging/preview branches which can be activated with feature flag. 
Get a mystery box on claim all successful.


prisma/migrations/20241118130418_add_mystery_box_prize_table/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+    "status" "BoxPrizeStatus" NOT NULL DEFAULT 'UnClaimed',
+    "size" "BoxSize" NOT NULL,
+    "prizeType" "BoxPrizeType" NOT NULL,
+    "amount" TEXT NOT NULL,
+    "claimHash" TEXT,
+    "rewardAmount" DECIMAL(65,30),
+    "claimFungibleTxId" TEXT,
+    "createdAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "updatedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "claimedAt" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
samott, line 42: Doesn't `claimedAt` need to be unset initially?

yash431garg: I think we can remove this as well as we have updatedAt already. What do you think?

samott: Yes, I think we only need one of those fields.

PR #682, opened 2024-11-20
fix: reveal all label

marvinmarnold:
Alternative to https://github.com/gator-labs/chomp/pull/681


marvinmarnold:
This doesn't seem to work either.

PR #685, opened 2024-11-21
feat: Add connect telegram and x functionality with hi-fi design

yash431garg:
- Description
User will be able to link/unlink x and telegram account from their connected wallet address simultaneously updating the database. 
Screenshot 2024-11-21 at 5 07 49 PM


app/actions/user.ts
```diff
@@ -0,0 +1,22 @@
+"use server";
```
marvinmarnold, line 1: Can you rename the file to addUserSocials?

app/components/SocialAuth/SocialAuth.tsx
```diff
@@ -0,0 +1,107 @@
...
+          onClick={async () => {
+            await linkSocialAccount(ProviderEnum.Twitter);
+          }}
+          className="bg-gray-700 hover:bg-gray-700"
+        >
+          <XIconWhite width={14} height={14} />
+          <p className="pl-4">Connect X</p>
+        </Button>
+      )}
+      {userTelegramInfo ? (
```
marvinmarnold, line 83: Doesn't look like this follows the AC "It should not be possible to disconnect TG if user signed up through the bot and there is no other wallet associated with the account."

marvinmarnold (changes requested):
Screenshot 2024-11-21 at 4 26 51 PM

I'm not able to load the login screen for some reason.

app/components/SocialAuth/SocialAuth.tsx
yash431garg: I think I missed that.

yash431garg:
> * Description
>   User will be able to link/unlink x and telegram account from their connected wallet address simultaneously updating the database.
> * Screen shots or recordings for UI changes
> 
> Screenshot 2024-11-21 at 5 07 49 PM

Done it was because of missing db migration. Now it should be working.

yash431garg:
> Done it was because of missing db migration. Now it should be working.

Done it was because of missing db migration. Now it should be working.

yash431garg:
> Screenshot 2024-11-21 at 4 26 51 PM
> 
> I'm not able to load the login screen for some reason.

Done it was because of missing db migration. Now it should be working.

marvinmarnold:
> Resolved it was because of missing db migration. Now it should be working.

Oh ok. This points out that we don't have a very clear process for how to handle migrations in staging. Some thoughts/questions

1. To fix the issue, did you manually run the migrations on staging? In some cases, that will cause the GH workflow to fail because it will try to reapply a duplicate change like "create index". It also gives us less confidence that the subsequent release to prod will work as expected.
2. An alternative approach is to first review the code, then merge into staging and test again there after migrations run. In this case, devs should just note that preview link wont work. 
3. We could also consider implementing some better dev infra. For example DB branching so there's a separate DB per preview link.

I think #2 is fine for now. It's pretty risk free to test things in staging.

 cc @samott

yash431garg:
> > Resolved it was because of missing db migration. Now it should be working.
> 
> Oh ok. This points out that we don't have a very clear process for how to handle migrations in staging. Some thoughts/questions
> 
> 1. To fix the issue, did you manually run the migrations on staging? In some cases, that will cause the GH workflow to fail because it will try to reapply a duplicate change like "create index". It also gives us less confidence that the subsequent release to prod will work as expected.
> 2. An alternative approach is to first review the code, then merge into staging and test again there after migrations run. In this case, devs should just note that preview link wont work.
> 3. We could also consider implementing some better dev infra. For example DB branching so there's a separate DB per preview link.
> 
> I think #2 is fine for now. It's pretty risk free to test things in staging.
> 
> cc @samott

Yes, I manually run the migrations. If we go with #2 Will we be testing the features in staging link and not in the preview branch.

marvinmarnold:
Old comment:

@yash431garg I'm still getting these errors

- Twitter error
Screenshot 2024-11-22 at 9 24 10 PM
- TG not working
Screenshot 2024-11-22 at 9 25 15 PM

marvinmarnold (changes requested):
@yash431garg this looks like it's working. but some follow ups:
- Can you add details in README or Notion about how to get X and Telegram keys necessary for setup? also about how to configure Dynamic correctly.
- Depending on the steps above, we may need to put one or both account linking options behind and env var feature flag.
- I'll have to test that tg can't be unlinked when sign up happens through bot after you merge into staging. 

> Yes, I manually run the migrations. If we go with https://github.com/gator-labs/chomp/pull/2 Will we be testing the features in staging link and not in the preview branch.

Yup. That's not ideal, but I think it's a good enough process for now.

yash431garg:
> @yash431garg this looks like it's working. but some follow ups:
> 
> * Can you add details in README or Notion about how to get X and Telegram keys necessary for setup? also about how to configure Dynamic correctly.
> * Depending on the steps above, we may need to put one or both account linking options behind and env var feature flag.
> * I'll have to test that tg can't be unlinked when sign up happens through bot after you merge into staging.
> 
> > Yes, I manually run the migrations. If we go with #2 Will we be testing the features in staging link and not in the preview branch.
> 
> Yup. That's not ideal, but I think it's a good enough process for now.

I have tested unlink telegram but dynamic throws an error and I have also added a custom error but let me know if you get any error.

marvinmarnold (approved):
I tested it and it works well. I also created a follow up task for us to discuss: https://linear.app/gator/issue/PROD-504/improve-infra-for-demo-environment

marvinmarnold:
Can you also test that tx signing works if you:

1. sign up with Phantom
2. connect Telegram
3. sign out
4. sign in with Telegram
5. try to reveal

It should still let you use Phantom burn. @neel-ds is this the edge case you were talking about, or did I misunderstand?

yash431garg:
> Can you also test that tx signing works if you:
> 
> 1. sign up with Phantom
> 2. connect Telegram
> 3. sign out
> 4. sign in with Telegram
> 5. try to reveal
> 
> It should still let you use Phantom burn. @neel-ds is this the edge case you were talking about, or did I misunderstand?

I have tested that you can check the comments on linear.

yash431garg:
> > Can you also test that tx signing works if you:
> > 
> > 1. sign up with Phantom
> > 2. connect Telegram
> > 3. sign out
> > 4. sign in with Telegram
> > 5. try to reveal
> > 
> > It should still let you use Phantom burn. @neel-ds is this the edge case you were talking about, or did I misunderstand?
> 
> I have tested that and you can check the comments on linear. https://linear.app/gator/issue/PROD-457/link-telegram-andor-x-account-from-profile#comment-5043b1db

Yihwan (approved):

PR #686, opened 2024-11-21
PROD-498 Better Handling Claim Tx Confirm Error

neel-ds:
- Description
Add extra params in ClaimError exception and handle the claim transaction confirmation error. Retry once after claim-tx-error then record the failure and send message to Slack

Manually throw error in claiming and verify from Slack or Sentry dashboard

NA


app/utils/claim.ts
```diff
@@ -95,13 +100,44 @@ export const sendBonk = async (
...
+    await CONNECTION.confirmTransaction(
+      {
+        signature,
+        ...blockhashResponse,
+      },
+      "confirmed",
+    );
+  } catch {
+    // If the claim transaction is not confirmed in first trial, try again with a latest blockhash
+    const blockhashResponse = await CONNECTION.getLatestBlockhash();
```
marvinmarnold, line 115: The original intention was 2 retries in addition to the original attempt. I don't feel super strongly about that. One retry could be OK.

But I also don't love that the retry code is duplicated instead of using a loop or a retry library.

app/utils/claim.ts
```diff
@@ -95,13 +100,44 @@ export const sendBonk = async (
...
+      await CONNECTION.confirmTransaction(
+        {
+          signature,
+          ...blockhashResponse,
+        },
+        "confirmed",
+      );
+    } catch {
+      // Record the error and send message to Slack
+      Sentry.captureMessage(
```
marvinmarnold, line 126: How about using Sentry.captureException? That way we can instantiate (but not throw) a custom error, defined in error.ts, like we do in other cases.

marvinmarnold (commented):
Some suggestions. But please prioritize the mystery box issue. Let's revisit this if you have time at the end of the sprint.

app/utils/claim.ts
neel-ds: Make sense, I will use a loop

app/utils/claim.ts
neel-ds: Sure

marvinmarnold:
@neel-ds merge conflicts

PR #687, opened 2024-11-21
PROD-495: Update sentry errors to display wallet address and username

samott:
Update sentry errors to display wallet address and username.

### DIFF
--- a/__tests__/actions/get_alltime-leaderboard.test.ts
+++ b/__tests__/actions/get_alltime-leaderboard.test.ts
@@ -92,6 +92,9 @@ describe("Get All-time leaderboard data", () => {
 
     let secondOrderOptionIndex = 0;
 
+    const qaData = [];
+    const fatlData = [];
+
     for (const user of users) {
       const selectedOption = questionOptions[Math.floor(Math.random() * 4)];
       const secondOrderOption = questionOptions[secondOrderOptionIndex];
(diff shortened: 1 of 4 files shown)


samott:
That last commit was related to PROD-486. I put it on the wrong branch.

__tests__/actions/get_alltime-leaderboard.test.ts
```diff
@@ -103,30 +106,30 @@ describe("Get All-time leaderboard data", () => {
...
+          change: Math.floor(Math.random() * 100),
+          type: TransactionLogType.AnswerQuestion,
         });
       }
+
       secondOrderOptionIndex =
         secondOrderOptionIndex === 3 ? 0 : secondOrderOptionIndex + 1;
     }
 
+    await prisma.questionAnswer.createMany({ data: qaData });
```
marvinmarnold, line 130: Nice optimization

app/actions/answer.ts
```diff
@@ -175,7 +176,7 @@ export async function answerQuestion(request: SaveQuestionRequest) {
     });
   } catch (error) {
     const answerError = new AnswerError(
-      `User with id: ${payload?.sub} is having trouble answering question with id: ${request.questionId}`,
+      `User with id: ${payload?.sub} (username: ${getUsername()}) is having trouble answering question with id: ${request.questionId}`,
```
marvinmarnold, line 179: I didn't realize we'd need to make another server request in order to fetch this data. That's not ideal because it could lead to a different error and further confuse things.

Probably best to just omit username from the logs, if it's not already available in the context.

What do you think?

samott: Yeah, agreed. I'll take out the username.

samott: I made that change. Maybe an idea for the future could be to have an admin panel endpoint that has all the user's details. A link could be included in the logs.

marvinmarnold (approved):

neel-ds (approved):
LGTM

PR #689, opened 2024-11-23
PROD-460 added question answer status check

AndrejRadomiljac:

- Description
- Link to the task -> https://linear.app/gator/issue/PROD-460/prevent-updating-question-once-answered. 
  After answering the question user should not be able to answer to the same question again. Right now, there is a check if someone triggers answer action and that question has already answered (status === 'submited') then we throw an error.
  Go to some deck in order to answer the question. Before answering open the network tab. Pay attention after submitting to the network tab and select the answer action. Copy the cUrl as it is provided in the screenshot:
Screenshot 2024-11-23 at 18 18 49
Paste that in your terminal and as a result there should be an error that says that user has already answered the question.
Screenshot 2024-11-23 at 18 18 38

### DIFF
--- a/app/actions/answer.ts
+++ b/app/actions/answer.ts
@@ -62,6 +62,24 @@ export async function answerQuestion(request: SaveQuestionRequest) {
       } as QuestionAnswer;
     });
 
+    const userQuestionAnswers = await prisma.questionAnswer.findMany({
+      where: {
+        questionOption: {
+          questionId: request.questionId,
+        },
+        userId,
+      },
+    });
+
+    if (
+      userQuestionAnswers.length &&
+      userQuestionAnswers.some((qa) => qa.status === AnswerStatus.Submitted)
+    ) {
+      throw new Error(
+        `User with id: ${payload?.sub} has already answered question with id: ${request.questionId}`,
+      );
+    }
+
     await prisma.$transaction(async (tx) => {
       await tx.questionAnswer.deleteMany({
         where: {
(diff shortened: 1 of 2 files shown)


marvinmarnold (changes requested):
Can you add a test case that `answerQuestion` won't create an answer twice? [Example](https://github.com/gator-labs/chomp/blob/main/__tests__/actions/deck/deck-editDeck.test.ts). 

Is there also some DB constraint that we can add so that QuestionAnswers are unique by (questionOptionId, userId)?

app/actions/answer.ts
```diff
@@ -62,6 +62,24 @@ export async function answerQuestion(request: SaveQuestionRequest) {
...
+
+    if (
+      userQuestionAnswers.length &&
+      userQuestionAnswers.some((qa) => qa.status === AnswerStatus.Submitted)
+    ) {
+      throw new Error(
+        `User with id: ${payload?.sub} has already answered question with id: ${request.questionId}`,
+      );
+    }
+
```
samott, line 82: Do you think it would be more efficient to integrate the check into the database query?

I also wonder if the guard should be inside the transaction below to avoid a race condition.

AndrejRadomiljac:
> Can you add a test case that `answerQuestion` won't create an answer twice? [Example](https://github.com/gator-labs/chomp/blob/main/__tests__/actions/deck/deck-editDeck.test.ts).
> 
> Is there also some DB constraint that we can add so that QuestionAnswers are unique by (questionOptionId, userId)?

Yes, I can add test case to confirm this. Unique constraint already exists.

marvinmarnold:
Thanks for that clarification. For me the most important piece is that `answerQuestion` is tested. And that we're sure there aren't any other code paths by which a user can answer. I defer to you all who are in the code more about when we should refactor.

marvinmarnold (approved):

samott (approved):

PR #690, opened 2024-11-25
PROD 477: Enable reward disperse of mystery box

neel-ds:
- Description
feat: Transfer reward amount to user wallet address and update the status of mystery box based on event. Record of MysteryBoxPrize get updated with `claimHash` and `claimedAt`

Trying claiming all questions and close the share link modal. It will automatically open a mystery box modal. Click on Open button and it will process the token transfer similar to claim func. 

NA


app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,238 @@
...
+        MysteryBoxPrize: {
+          select: {
+            id: true,
+          },
+        },
+      },
+    });
+    return res.id;
+  } catch (e) {
+    console.log(e);
```
samott, line 75: return `null` for consistency (does the upstream code actually handle the failure case?)

app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,238 @@
...
+      data: {
+        status: EMysteryBoxStatus.Unopened,
+      },
+    });
+    const claimMysteryBoxError = new ClaimMysteryBoxError(
+      `User with id: ${payload.sub} is having trouble claiming for Mystery Box: ${mysteryBoxId}`,
+      { cause: e },
+    );
+    Sentry.captureException(claimMysteryBoxError);
+    release();
```
samott, line 183: An exception further up in the catch handler would cause `release()` not to be called (not good)

app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  let sendTx: string | null = null;
+  if (rewardAmount > 0) {
+    sendTx = await sendBonk(
+      treasuryWallet,
+      new PublicKey(address),
+      Math.round(rewardAmount * 10 ** 5),
+    );
+  }
```
samott, line 235: I would declare the variable in the if statement (and `return null`) at the end of the function

app/actions/mysteryBox.ts
neel-ds: Make sense, I will put in finally statement

app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,233 @@
...
+    await prisma.mysteryBox.update({
+      where: {
+        id: mysteryBoxId,
+      },
+      data: {
+        status: EMysteryBoxStatus.Unopened,
+      },
+    });
+    const claimMysteryBoxError = new ClaimMysteryBoxError(
+      `User with id: ${payload.sub} is having trouble claiming for Mystery Box: ${mysteryBoxId}`,
```
samott, line 172: Can you add the wallet address to the log also?

app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,233 @@
...
+    if (!userWallet) {
+      return;
+    }
+
+    const sendTx = await handleSendBonk(
+      Number(reward?.MysteryBoxPrize[0]?.amount),
+      userWallet.address,
+    );
+
+    if (!sendTx) throw new Error("Send tx is missing");
```
marvinmarnold, line 115: Better error message. Including user id, wallet address, and clarify that this is mystery box, not regular claim.

app/actions/mysteryBox.ts
```diff
@@ -0,0 +1,233 @@
...
+      { cause: e },
+    );
+    Sentry.captureException(claimMysteryBoxError);
+    throw e;
+  } finally {
+    release();
+  }
+}
+
+export async function handleSendBonk(rewardAmount: number, address: string) {
```
marvinmarnold, line 182: Move treasury warning logic into sendBonk. Remove code duplication from pre-existing handleSendBonk.

PR #695, opened 2024-11-26
fix: mystery box without token transfer

yash431garg:


lib/error.ts
```diff
@@ -50,3 +50,9 @@ export class ClaimMysteryBoxError extends Error {
     this.name = "ClaimMysteryBoxError";
   }
 }
+export class MysteryBoxError extends Error {
```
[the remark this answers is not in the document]
neel-ds: Resolved in recent commit

app/components/ClaimShareDrawer/ClaimShareDrawer.tsx
```diff
@@ -45,8 +45,12 @@ const ClaimShareDrawer = ({
 
   useEffect(() => {
     const fetchLinkPreview = async () => {
-      const linkPreview = await getLinkPreview(claimUrl);
-      setOgImageUrl((linkPreview as { images: string[] }).images[0]);
+      try {
+        const linkPreview = await getLinkPreview(claimUrl);
+        setOgImageUrl((linkPreview as { images: string[] }).images[0]);
+      } catch {
+        console.log("Failed to fetch share claim all preview");
```
marvinmarnold, line 52: Should this go to sentry or is that happening elsewhere?

yash431garg: No, It isn't going to Sentry sometime the above function fail and throw error to client. It is to prevent that for now. But can be sent to Sentry.

cc: @neel-ds what do you think?

PR #704, opened 2024-11-28
PROD-512 added error state for answering revealed questions

AndrejRadomiljac:
Link to the Linear https://linear.app/gator/issue/PROD-512/ensure-its-not-possible-to-answer-questions-that-have-already-been
- Description
- User was able to answer the revealed question. I've added condition that excludes those questions and throws error in that case. I've also added tests, but after latest updates on staging I am not able to run any test. I am creating PR without tests until we figure out what is causing the issue.


__tests__/actions/answer/answer-revealed-question.test.ts
```diff
@@ -0,0 +1,260 @@
...
+    const request: SaveQuestionRequest = {
+      questionId,
+      questionOptionId,
+      percentageGiven: 50,
+      timeToAnswerInMiliseconds: 5000,
+    };
+
+    await answerQuestion(request);
+
+    expect(deleteManyMock).toHaveBeenCalledWith({
```
marvinmarnold, line 148: Generally, I think `toHaveBeenCalled` should be avoided. Tests should be oriented around testing real user flows. In this case, I would not mock the DB connection. You can then make assertions about the entry in the DB.

This won't be a true user test. But it will give us more confidence that the underlying behavior works and won't be implementation specific.

__tests__/actions/answer/answer-revealed-question.test.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    try {
+      await answerQuestion(request);
+    } catch (error: any) {
+      expect(error.message).toBe(
+        `User with id: ${userId} has already answered question with id: ${questionId}`,
+      );
+    }
+
+    expect(prisma.questionAnswer.createMany).not.toHaveBeenCalled();
```
marvinmarnold, line 258: Similar to my other comment. I would assert that DB does not have any new entries. Instead of asserting that createMany wasn't called.

marvinmarnold (changes requested):

__tests__/actions/answer/answer-revealed-question.test.ts
AndrejRadomiljac: expect(deleteManyMock).toHaveBeenCalledWith
expect(deleteManyMock).toHaveBeenCalledWith
are leftovers. My bad. But generally, generally this test is E2E test where I send data to the API endpoint and check the response. In this case we expect error, due the revealAt date. That's the reason why I intentionally inserted DB records. There is no other way to test, because you need to be aware of question ids that should be part of test. Unit test for this does not make sense since you we are trying to test the ORM gte method. @marvinmarnold

AndrejRadomiljac:
I've mixed somehow test for answer revealed and answer twice. Now everything is up to date.

__tests__/actions/answer/answer--revealed-question.test.ts
```diff
@@ -0,0 +1,159 @@
...
+      questionId: currentQuestionId,
+      questionOptionId: currentQuestionOptions[0].id,
+      percentageGiven: 50,
+      timeToAnswerInMiliseconds: 5000,
+    };
+
+    try {
+      await answerQuestion(request); // Call the method under test
+    } catch (error: any) {
+      expect(error.message).toBe(
```
marvinmarnold, line 154: > are leftovers. My bad. But generally, generally this test is E2E test

Gotcha. I think this test could still be stronger if you asserted that the only answers that exist in the DB are the ones created in setup. Otherwise it's possible that the API is just returning an exception, but actually a new answer was actually inserted.

marvinmarnold (dismissed):

__tests__/actions/answer/answer--revealed-question.test.ts
```diff
@@ -0,0 +1,159 @@
+import { SaveQuestionRequest, answerQuestion } from "@/app/actions/answer";
```
marvinmarnold, line 1: This filename also has an extra `-`

marvinmarnold:
@AndrejRadomiljac are you going to do follow up on this or are you waiting for reviews?

danijel-radomiljac (dismissed):

PR #706, opened 2024-11-29
test: add unit tests for date utilities and streak end date logic

yash431garg:


__tests__/utils/date.test.ts
```diff
@@ -0,0 +1,222 @@
...
+describe("Time-Based Chomp Actions", () => {
+  const currentDate = new Date();
+
+  describe("Daily Leaderboard Refresh", () => {
+    test("Refreshes daily at midnight UTC", () => {
+      const currentDate = new Date();
+
+      // Get calculated end date from utility
+      const { endOfTheDay: calculatedEndDate } =
+        getStartAndEndOfDay(currentDate);
```
marvinmarnold, line 38: It doesn't hurt to test these lower level methods. But can you also test the entry points? Like [getLeaderboard](https://github.com/gator-labs/chomp/blob/05abd6a52463fc59abcf1bca8dc8e7b8b2511a4d/app/actions/leaderboard.ts#L133). 

You are already testing some top level entry points like `getUsersLatestStreak`, which is good. But make sure all the interfaces that the frontend interacts with have a test.

marvinmarnold (changes requested):

__tests__/utils/date.test.ts
yash431garg: Do you mean to test top level function like getLeaderboard but only test that the date matches the required end time.

marvinmarnold: Yes. Whatever actions the frontend invokes to get streaks and leaderboard. Test that output is as expected based on date and time.

danijel-radomiljac (approved):

PR #709, opened 2024-12-02
fix: add query to remove old transaction log in reveal

yash431garg:
- Description
Added query to remove fungible transaction before completing the reveal query. 
- Steps to test
Try to reveal and reveal all.

### DIFF
--- a/app/actions/chompResult.ts
+++ b/app/actions/chompResult.ts
@@ -138,16 +138,12 @@ export async function revealQuestions(
 
   const revealableQuestionIds = revealableQuestions.map((q) => q.id);
 
-  let questionRewards: {
-    questionId: number;
-    rewardAmount: number;
-    revealAmount: number;
-  }[];
-  let revealPoints;
+  const questionRewards = await calculateReward(
+    payload.sub,
+    revealableQuestionIds,
+  );
 
-  questionRewards = await calculateReward(payload.sub, revealableQuestionIds);
-
-  revealPoints = await calculateRevealPoints(questionRewards);
+  const revealPoints = await calculateRevealPoints(questionRewards);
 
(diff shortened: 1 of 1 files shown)


app/actions/chompResult.ts
```diff
@@ -205,7 +207,13 @@ export async function revealQuestions(
         questionId: revealPointsTx.questionId,
       })),
     });
-  });
+  } catch (error) {
+    const revealError = new RevealError(
+      `Reveal point error for User id: ${payload?.sub} and ${questionIds}`,
```
marvinmarnold, line 212: Can you make the message more actionable:

Reveal warning for user ${payload?.sub} and ${questionIds}. Reveal will continue successfully, but unable to create new FATL. Existing FATL ${}. New FATL ${}

You'll need to make an additional query in order to fetch the existing FATL, only in the error case.

marvinmarnold:
> We may need another ticket to address the handling of calculateRewards error and refactoring the current implementation.

Let's use this to track follow up work: https://linear.app/gator/issue/PROD-525/burn-txs-not-succeeding-leads-to-invalid-burntxhash-in-chompresult-and#comment-52ed090c

app/actions/chompResult.ts
```diff
@@ -195,17 +195,46 @@ export async function revealQuestions(
...
         type: revealPointsTx.type,
         change: revealPointsTx.amount,
         userId: payload.sub,
         questionId: revealPointsTx.questionId,
       })),
     });
-  });
+  } catch (error) {
+    const questionIds = questionRewards.map((item) => item.questionId);
+    const ExistingFatl = await prisma.fungibleAssetTransactionLog.findMany({
```
marvinmarnold, line 212: Should be camel case existingFatl

PR #713, opened 2024-12-03
PROD-528: Mystery box improvements

samott:
- Adds a scrollbar so the content is accessible on small screens.
- Add space between the button and total winnings.
- Fixes the background of the treasure chest image.


components/MysteryBox/MysteryBoxOverlay.tsx
```diff
@@ -6,7 +6,7 @@ type MysteryBoxOverlayProps = PropsWithChildren;
 function MysteryBoxOverlay({ children }: MysteryBoxOverlayProps) {
   return (
     <RemoveScroll>
-      <div className="fixed inset-0 z-100 bg-black/90 mt-[3em] z-0 flex justify-center">
+      <div className="fixed inset-0 z-100 bg-black/90 mt-[3em] z-0 flex justify-center overflow-scroll mb-[5em]">
```
marvinmarnold, line 9: Can you use overflow-y-scroll instead? Otherwise, on some devices you may wind up with an unnecessary horizontal scroll bar.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -134,8 +134,8 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
     <>
       <MysteryBoxOverlay>
         {!box && (
-          <div className="fixed z-100 flex flex-col items-center justify-center content-between h-[calc(100%_-_5.5em)] absolute">
-            <div className="w-full grow-[2] flex flex-col gap-8 justify-center">
+          <div className="fixed z-100 flex flex-col items-center m-auto content-between h-full absolute">
+            <div className="w-full flex flex-col gap-8 justify-center mt-[2em]">
```
marvinmarnold, line 138: General question for @gator-labs/devs. When do we use em, rem, px, or tailwind number (eg `my-20`)?

marvinmarnold (changes requested):

components/MysteryBox/MysteryBoxOverlay.tsx
samott: No problem; I think the `y` got lost.

components/MysteryBox/MysteryBox.tsx
neel-ds: We use em for margin or padding, rem for font size or line height and px for icon size or border width. Tailwind numbers are abstracted pre-defined values. Generally, it is used for height or width -- margin or padding.

yash431garg: We have different units like you mentioned em, rem can be vary depending on the parent/child width, height and other properties. But px is more independent and fix. As we use tailwind in our project. I suggest to use tailwind numbers as they will follow the convention and easy to change and understand.

yash431garg: I will suggest to use mt-2,3 depending upon the figma design. You can check exact number by using dev mode in figma and refer to the link below for the exact number according to tailwind css. 

https://tailwindcss.com/docs/margin

samott: I've made some refinements. Thanks for the feedback.

components/MysteryBox/MysteryBoxAmount.tsx
```diff
@@ -19,7 +19,7 @@ function MysteryBoxAmount({
   return (
     <div
       className={cn(
-        "text-sm font-bold p-1 px-2 rounded-[2em] flex items-center justify-center text-center",
+        "text-xs font-bold p-1 px-3 rounded-[2em] flex items-center justify-center text-center",
```
yash431garg, line 22: We can use rounded-md/sm something like that depending upon the need instead of em.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -134,105 +134,93 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
             </div>
 
-            <Image
-              src={IMAGES[image]}
-              alt="Treasure Chest"
-              title="Treasure Chest"
-              className="my-16 cursor-pointer w-[212px]"
-              onClick={openBox}
-            />
+            <div className="py-[35px]">
```
yash431garg, line 146: We can use py-9

components/MysteryBox/MysteryBox.tsx
```diff
@@ -134,105 +134,93 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
+              <Image
+                src={IMAGES[image]}
+                alt="Treasure Chest"
+                title="Treasure Chest"
+                className="cursor-pointer w-[182px] h-[182px]"
```
yash431garg, line 151: @neel-ds can we use tailwind here as well or custom size is fine?

samott: > @neel-ds can we use tailwind here as well or custom size is fine?

I'll leave this one as-is in any case as the image will be replaced soon,

neel-ds (approved):
Noticed some leftover console.log statements in the mystery box tests. It could be from previous debugging. Can you remove them to keep the logs clean?

samott:
I merged in PROD-538. Please re-review.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -124,115 +138,105 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
-          <div className="fixed z-100 flex flex-col items-center justify-center content-between h-[calc(100%_-_5.5em)] absolute">
-            <div className="w-full grow-[2] flex flex-col gap-8 justify-center">
-              <h1 className={`text-chomp-green-light text-2xl font-bold`}>
-                You earned a mystery box!
-              </h1>
-
-              <div className="text-center text-sm">
-                {buildMessage(OPEN_MESSAGES[message].subText)}
-              </div>
+          <div className="fixed z-100 flex flex-col items-center m-auto justify-between h-full absolute pt-10 pb-10">
```
neel-ds, line 153: I noticed that absolute might not be needed here since the position is already set to fixed.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -124,115 +138,105 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
         )}
 
         {box && (
-          <div className="fixed z-100 flex flex-col items-center justify-center content-between h-[calc(100%_-_5.5em)] absolute">
-            <div className="w-full grow-[2] flex flex-col gap-8 justify-center">
-              <h1 className={`text-chomp-green-light text-2xl font-bold`}>
-                CHOMP, CHOMP HOORAY!
-              </h1>
-            </div>
+          <div className="fixed z-100 flex flex-col items-center m-auto justify-between h-full absolute pt-10 pb-10">
```
neel-ds, line 192: It seems the position is defined twice as mentioned above.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -124,115 +138,105 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
-              <div className="flex items-start items-center w-full content-center justify-center gap-2 mt-[1em]">
-                <span className="text-sm grow-3">Total $BONK won to date</span>
-                <MysteryBoxAmount type="tokens" amount={box.totalBonkWon} />
+            {box.creditsReceived > 0 && (
+              <div className="text-sm text-center">
+                Learn more about credits
               </div>
+            )}
+
+            <div className="flex items-start items-center w-full content-center justify-center gap-2 my-2">
```
neel-ds, line 224: We plan to keep items-center as per the design. Could you confirm if that’s okay? If so, you can remove items-start.

marvinmarnold (approved):
Behavior looks good to me. To continue the Linear convo, can you clarify why the chest still has some weird specs and rough edges? Are they really part of the image in Figma? I didn't see when I pasted into gimp.

Screenshot 2024-12-05 at 11 45 26 AM

samott:
> Behavior looks good to me. To continue the Linear convo, can you clarify why the chest still has some weird specs and rough edges? Are they really part of the image in Figma? I didn't see when I pasted into gimp.
> 
> Screenshot 2024-12-05 at 11 45 26 AM

No they are not part of the image; I did a second export (as PNG) which still has the artifacts. I also removed the margin which might have also introduced noise. Anyway, the solution is using the source SVG.

samott:
> > Anyway, the solution is using the source SVG.
> 
> Should that be part of this PR? Or maybe it's moot because of the animation work that is in flight? I haven't fully looked at that yet.

I left it out since the other PR is in flight (seems like it needs more work though...)

neel-ds (approved):

PR #714, opened 2024-12-03
Fix an issue with claim URL generation.

samott:
Resolves an issue with the claim share feature (`window` is not available on the server so we need an effect).

### DIFF
--- a/lib/urls.ts
+++ b/lib/urls.ts
@@ -9,11 +9,11 @@ export const ANSWER_PATH = `${HOME_PATH}/answer`;
 export const getOgShareClaimAllPath = (startOfTxHash: string) =>
   `/api/og/share-claim-all?startOfTxHash=${startOfTxHash}`;
 export const getClaimAllShareUrl = (startOfTxHash: string) => {
-  return `${window?.location?.origin ?? ""}/a/${startOfTxHash}`;
+  return `${window.location.origin}/a/${startOfTxHash}`;
 };
 
 export const getClaimSingleShareUrl = (txHashAndQId: string) => {
-  return `${window?.location?.origin ?? ""}/s/${txHashAndQId}`;
+  return `${window.location.origin}/s/${txHashAndQId}`;
 };
 export const getOgShareClaimSinglePath = (txHashAndQId: string) =>
   `/api/og/share-claim-single?txHashAndQId=${txHashAndQId}`;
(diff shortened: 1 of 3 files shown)


marvinmarnold (commented):
Are there logs on Sentry or somewhere you can point to where we hit this error? Some thoughts:

- I'm surprised the server was trying to access it because it looks like the place it's called from has "use client" up top
- Maybe we should pull this helper method into a different file with "use client" if that's how it's used
- Could use the env var NEXT_PUBLIC_APP_URL which is already set instead of referencing window.

samott:
> Are there logs on Sentry or somewhere you can point to where we hit this error? Some thoughts:
> 
>     * I'm surprised the server was trying to access it because it looks like the place it's called from has "use client" up top
> 
>     * Maybe we should pull this helper method into a different file with "use client" if that's how it's used
>     * Could use the env var NEXT_PUBLIC_APP_URL which is already set instead of referencing window.

Client components still get pre-rendered on the server.

Error seen here: https://gator-labs.sentry.io/issues/6083619368/?environment=vercel-preview&project=4507686429130752

I saw it this morning locally - it blocked a mystery box open and there were a couple in the logs from when we were testing the other day.

Using the env var also works.

marvinmarnold (dismissed):
Thanks for the clarification.

lib/urls.ts
```diff
@@ -9,11 +9,11 @@ export const ANSWER_PATH = `${HOME_PATH}/answer`;
 export const getOgShareClaimAllPath = (startOfTxHash: string) =>
   `/api/og/share-claim-all?startOfTxHash=${startOfTxHash}`;
 export const getClaimAllShareUrl = (startOfTxHash: string) => {
-  return `${window.location.origin}/a/${startOfTxHash}`;
+  return `${window?.location?.origin ?? ""}/a/${startOfTxHash}`;
```
yash431garg, line 12: I think if we do it like this it can pass a url like this ""/a/hash. How about returning a null if it is not defined for the whole url.

samott: It's kinda weird returning null though -- it'd probably be better as a hook in that case.

I think I will go with @marvinmarnold's idea of using the environment variable as it removes the burden on the caller and lets us control the hostname that is used.

marvinmarnold (approved):

danijel-radomiljac (approved):

PR #716, opened 2024-12-05
PROD-521/replace-placeholder-treasure-box-with-bonk-gif-animation

danijel-radomiljac:


components/MysteryBox/MysteryBox.tsx
samott, line 1: The animation doesn't have time to complete if the box opens quickly (which will be the case after PROD-538).

Is this PR complete? The design spec had a multi-stage process, with the controls hiding, box opening then shrinking into place to show the reward amounts.

danijel-radomiljac: Screenshot 2024-12-05 at 12 38 12
I cannot open video on the right. But I am not sure what do you mean by multi-stage process, with the controls hiding etc.

samott: It's in the video. The image expands then shrinks (so it is where the treasure box image is on the screen that shows the winnings). Are you able to request access to the video?

marvinmarnold (commented):
Agree with Shaun's comments

components/MysteryBox/MysteryBoxOverlay.tsx
```diff
@@ -6,7 +6,7 @@ type MysteryBoxOverlayProps = PropsWithChildren;
 function MysteryBoxOverlay({ children }: MysteryBoxOverlayProps) {
   return (
     <RemoveScroll>
-      <div className="fixed inset-0 z-100 bg-black/90 mt-10 z-0 flex justify-center overflow-y-scroll mb-20">
+      <div className="fixed inset-0 z-[999] bg-black/90 flex justify-center">
```
samott, line 9: The problem with this change is it means the header/footer aren't visible as they are in the design.

samott: Also, the overflow scroll was removed so the bottom of the content can be cut off if the screen is small enough.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -140,106 +149,138 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
+                <Image
+                  src={Coins.src}
+                  alt="coins"
+                  fill
+                  className="object-cover"
+                />
+              </div>
+              <div className="w-full bg-chomp-orange-dark text-xs rounded-[56px] py-2 px-4">
+                {bonkReceived.toLocaleString("en-US")} BONK
+              </div>
```
samott, line 240: There is a component for the amount (`<MysteryBoxAmount />`) labels, btw.

components/MysteryBox/MysteryBox.tsx
```diff
@@ -140,106 +149,138 @@ function MysteryBox({ isOpen, closeBoxDialog, mysteryBoxId }: MysteryBoxProps) {
...
+                    : status === "Closing"
+                      ? "0.8"
+                      : "1",
+                zIndex: 999,
+                transform: `translateY(${status === "Closing" ? -118 : -70}%) translateX(-43%)`,
+              }}
+              className="absolute top-1/2 left-1/2"
+              onClick={openBox}
+              disabled={isSubmitting || !!box}
             />
```
samott, line 221: The image is offset to the right at the end of the animation (maybe something for the designers to fix...)

components/MysteryBox/MysteryBoxOverlay.tsx
danijel-radomiljac: In my opinion header/footer should not visible. Because user should not be able to exit the modal while animation is in progress. @gatorthana @marvinmarnold

marvinmarnold: @danijel-radomiljac can you show header/footer before animation, but hide after the user clicks redeem/animation starts?

marvinmarnold: Ideally header/footer would only be hidden after claim starts. But assuming that's tricky, we can just hide for now.

samott (approved):

marvinmarnold:
Merging this and will make follow up comments in Linear

PR #719, opened 2024-12-06
PROD-523: Add Mysterybox Allowlist

neel-ds:
- Description

It replaces the Mystery Box FF with Allowlist. It contains schema update to support batch of different campaigns. 

Use the script to add your address in allowlist and claim the rewards. You will see the mystery box.

NA


app/actions/claim.ts
```diff
@@ -257,3 +258,26 @@ export async function claimQuestions(questionIds: number[]) {
     throw e;
   }
 }
+
+export async function isUserInAllowlist(): Promise<boolean> {
```
samott, line 262: Actions should get their own file, but this is a utility so should go in `lib` or `utils` (outside `app`)

app/actions/mysteryBox/reward.ts
```diff
@@ -47,6 +49,7 @@ export async function rewardMysteryBox({
             data: questionIds.map((questionId) => ({
               questionId,
               triggerType,
+              mysteryBoxAllowlistId: user?.wallets[0].address ?? "",
```
samott, line 52: As above, we probably want to check all wallets.

app/components/ClaimShareDrawer/ClaimShareDrawer.tsx
```diff
@@ -66,11 +68,19 @@ const ClaimShareDrawer = ({
...
-  const FF_MYSTERY_BOX = process.env.NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL;
+
+  const checkAllowlist = async () => {
+    const isAllowlisted = await isUserInAllowlist();
+    setIsUserWhitelisted(isAllowlisted);
+  };
+
+  useEffect(() => {
+    checkAllowlist();
+  }, []);
```
samott, line 79: Seems like we only need to check this if `mysteryBoxId` is defined.

samott: In fact, `mysteryBoxId` won't ever be defined if the user isn't whitelisted, right? Maybe we don't need this check.

app/actions/mysteryBox/reward.ts
neel-ds: Here, it adds current user wallet while creating mystery box & mystery box trigger.

samott (approved):

scripts/addAllowlistUsers.ts
```diff
@@ -0,0 +1,55 @@
...
+  const csvFilePath = path.resolve(__dirname, "allowlist.csv");
+  const csv = fs.readFileSync(csvFilePath, "utf8");
+  const rows = csv.split(/\r?\n/);
+
+  await prisma.mysteryBoxAllowlist.createMany({
+    data: rows.map((address: string) => ({
+      address: address.trim(),
+      tags,
+    })),
+    skipDuplicates: true,
```
marvinmarnold, line 39: This means that an address will only ever have one tag, since we don't append new ones. Not a blocker because it's a small detail. We'll also always have the original CSVs as reference.

yarn.lock
marvinmarnold, line 1: Remove this file from PR and merge in changes from main.

app/actions/claim.ts
```diff
@@ -91,9 +92,9 @@ export async function claimAllAvailable() {
 
   if (!claimableQuestionIds.length) throw new Error("No claimable questions");
 
-  const FF_MYSTERY_BOX = process.env.NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL;
+  const isAllowlisted = await isUserInAllowlist();
```
marvinmarnold, line 95: Allow list only relevant for CHOMPmas mystery box. ClaimAllCompleted does not require presence in allowlist.

app/components/ClaimShareDrawer/ClaimShareDrawer.tsx
```diff
@@ -66,11 +66,10 @@ const ClaimShareDrawer = ({
 
     if (!!copyUrl) fetchLinkPreview();
   }, [isOpen, copyUrl]);
-  const FF_MYSTERY_BOX = process.env.NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL;
 
   return (
     <>
-      {FF_MYSTERY_BOX && showMysteryBox && mysteryBoxId ? (
+      {showMysteryBox && mysteryBoxId ? (
```
marvinmarnold, line 72: FF_MYSTERY_BOX should still be used.

lib/mysteryBox.ts
```diff
@@ -43,3 +45,28 @@ export async function sendBonkFromTreasury(
...
+      where: {
+        address: {
+          in: user?.wallets.map((wallet) => wallet.address) || [],
+        },
+      },
+    });
+
+    return !!allowlist;
+  } catch (e) {
+    console.log("Error in isUserInAllowlist", e);
```
marvinmarnold, line 69: Should console.error so that it's automatically sent to Sentry

README.md
```diff
@@ -175,3 +175,18 @@ yarn dev:restore-db
 ```
 
 If you see a “restore successful” message, you can safely ignore some errors. After verifying the restored database, don’t forget to remove the backup file.
+
```
marvinmarnold, line 178: Thank you for adding docs!

marvinmarnold (changes requested):

scripts/addAllowlistUsers.ts
neel-ds: Yes, Do you want me to change the datatype -> String[]?

app/actions/claim.ts
neel-ds: This condition is used to create mysteryBox once ClaimAllCompleted. If user doesn't exist in allowlist then it won't even create mysteryBox.

marvinmarnold:
@samott will merge your work into his and open a new pr

PR #720, opened 2024-12-08
Prod 532: rpc tx confirmation backend

yash431garg:
- Description
1. Reveal refactoring
2. Remove tx confirmation from frontend
3. Update burntx creation
4. Send sentry error if tx not validated.


app/utils/solana.ts
```diff
@@ -79,11 +81,15 @@ export const genBonkBurnTx = async (
 
   tx.add(modifyComputeUnits);
```
marvinmarnold, line 82: const burnTx = new Transaction();
burnTx.add(modifyComputeUnits);
burnTx.add(addPriorityFee);
burnTx.add(burnTxInstruction);

app/hooks/useReveal.ts
```diff
@@ -314,55 +267,55 @@ export function useReveal({ wallet, address, bonkBalance }: UseRevealProps) {
...
+          //       tags: {
+          //         category: "reveal-tx-confirmation-error",
+          //       },
+          //       extra: {
+          //         transactionHash: signature,
+          //       },
+          //     });
+          //   }
+
+          //   await deleteQuestionChompResults(pendingChompResultIds);
```
marvinmarnold, line 317: Do all this
1. can be deleted
2. copied and moved to backend function
3. Check for expired blockhash specifically and delete ChompResults in that case only. Otherwise create ChompResult in state `ValidationFailed`.

app/actions/chompResult.ts
```diff
@@ -346,6 +345,7 @@ export async function getUsersPendingChompResult(questionIds: number[]) {
   });
 }
 
+// tx validation with 5 retries with a delay of 1s
```
marvinmarnold, line 348: Call confirmTransaction first. It will either return:
1. Success -> Continue on to getParsedTransaction
2. Fail with expired blockhash -> Delete ChompResult. Do not continue.
3. Fail with unknown error -> Continue on to getParsedTransaction

Call getParsedTransaction
1. Success -> Update record to Completed
2. Failure -> Update record to VerificationFailed

yash431garg: Check for expired blockhash specifically and delete ChompResults in that case only. Otherwise create ChompResult in state ValidationFailed.

app/utils/solana.ts
```diff
@@ -68,18 +77,10 @@ export const genBonkBurnTx = async (
...
-  const modifyComputeUnits = ComputeBudgetProgram.setComputeUnitLimit({
-    units: Math.round(computeUnitFix * 1.1),
-  });
-
   const addPriorityFee = ComputeBudgetProgram.setComputeUnitPrice({
     microLamports: Math.round(estimateFee?.result?.priorityFeeLevels?.high),
   });
 
-  tx.add(modifyComputeUnits);
   tx.add(addPriorityFee);
```
marvinmarnold, line 84: Should we also move this up to before where we add the burnTx?

app/actions/chompResult.ts
yash431garg: I am not marking chompResult as ValidationFailed here because in the cron job we will fetch the pending requests and if cron still not able to verify the tx than we will mark it is as Failed it will help in double verification this way.

marvinmarnold (changes requested):
Reveal all not working: https://linear.app/gator/issue/PROD-532/reveal-move-rpc-tx-confirmation-to-backend#comment-538ab9ac

app/actions/validateBonkBurn.ts
marvinmarnold, line 1: Move into lib. And can remove SECRET.

app/api/process-pending-transaction/route.ts
```diff
@@ -0,0 +1,130 @@
...
+import prisma from "@/app/services/prisma";
+import { calculateReward } from "@/app/utils/algo";
+import { RevealConfirmationError } from "@/lib/error";
+import { ResultType, TransactionStatus } from "@prisma/client";
+import * as Sentry from "@sentry/nextjs";
+import { startOfDay, sub } from "date-fns";
+
+export async function GET(request: Request) {
+  const authHeader = request.headers.get("authorization");
+  const SECRET = process.env.CRON_SECRET || "";
```
marvinmarnold, line 11: SECRET should be lower case

lib/validateBonkBurn.ts
```diff
@@ -0,0 +1,47 @@
...
+    return false;
+  }
+
+  const burnInstruction = transaction.transaction.message.instructions.find(
+    (instruction) =>
+      "parsed" in instruction &&
+      instruction.parsed.type === "burnChecked" &&
+      wallets.includes(instruction.parsed.info.authority) &&
+      instruction.parsed.info.mint ===
+        "DezXAZ8z7PnrnRJjz3wXBoRgixCa6xjnB7YaB1pPB263",
```
marvinmarnold, line 39: Use constant for this

PR #721, opened 2024-12-09
PROD-510: Protect mystery box endpoint

samott:
Remove `rewardMysteryBox()` as a server action endpoint.


lib/mysteryBox.ts
```diff
@@ -43,3 +53,55 @@ export async function sendBonkFromTreasury(
...
+export async function rewardMysteryBox(
+  userId: string,
+  triggerType: EBoxTriggerType,
+  questionIds: number[],
+) {
+  try {
+    const calculatedReward = await calculateMysteryBoxReward(
+      MysteryBoxEventsType.CLAIM_ALL_COMPLETED,
+    );
+    const tokenAddress = process.env.NEXT_PUBLIC_BONK_ADDRESS ?? "";
```
marvinmarnold, line 73: Perhaps we should check that calculatedReward.bonk is defined. Otherwise error out. The code you have now should work for bonk only, especially because reward should always be defined. But just an extra level of sanity check.

marvinmarnold (changes requested):
Can you remove changes to yarn.lock?

marvinmarnold:
@samott yarn.lock is stale again

PR #729, opened 2024-12-11
PROD 541: Chompmas streaks mystery boxes

samott:
Implements whitelisted Chompmas streaks mystery boxes.

Requires the mechanism update to function.


.env.example
```diff
@@ -37,8 +37,7 @@ OVERRIDE_USER_ID=
...
 
-# mystery box feature flag
-NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL=
-
 # bot feature flag
-FF_CHOMP_BOT=
\ No newline at end of file
+FF_CHOMP_BOT=
+
+CHOMPMAS_MIN_STREAK=7
```
marvinmarnold, line 43: Add to Vercel and .env.example

marvinmarnold:
I set CHOMPMAS_MIN_STREAK=1 in preview, but I'm not getting a mystery box
Screenshot 2024-12-11 at 4 23 58 PM

marvinmarnold:
Oh I need to add myself to allowlist. I'll try that.

README.md
```diff
@@ -175,3 +175,18 @@ yarn dev:restore-db
...
+
+## Add Users in MysteryBox Allowlist
+
+1. Add a csv file in the scripts folder (only addresses, no header)
+2. Make sure the name of file is `allowlist.csv`
+3. Verify the DATABASE_URL env
+
+**Run the following command**
+```sh
+yarn dev:add-mystery-box
```
marvinmarnold, line 187: package.json uses: dev:add-allowlist-users

samott:
We also need the mechanism-engine update.

marvinmarnold:
> We also need the mechanism-engine update.

We also need to run migrations for this to work. For now I just want to be able to test in staging so I want to merge. Can you:
1. Use the existing event type for claim all instead of chompas?
2. Double check that this code will not impact prod if merged (ie hidden behind FF feature flag)

samott:
> > We also need the mechanism-engine update.
> 
> We also need to run migrations for this to work. For now I just want to be able to test in staging so I want to merge. Can you:
> 
>     1. Use the existing event type for claim all instead of chompas?

Done

>     2. Double check that this code will not impact prod if merged (ie hidden behind FF feature flag)

I added an extra flag. Everything looks good with it switched off.

PR #737, opened 2024-12-13
PROD-524/create-streaks-flow-mvp-for-chompmas 

danijel-radomiljac:


app/components/LatestStreakBox/LatestStreakBox.tsx
```diff
@@ -24,6 +25,43 @@ const LatestStreakBox = ({
     setMysteryBoxClosed(false);
   }, [mysteryBoxId]);
 
+  if (!!mysteryBoxId && latestStreak >= 7)
```
samott, line 28: We have an env var (CHOMPMAS_MIN_STREAK) for this value.

Also, there is an edge case where the user hits 7 days, gets a box but doesn't claim, then loses the streak. They should still be able to click to see retrieve box.

danijel-radomiljac: I will add that var. 

I did not know about that case it was not specified in a ticket.

yash431garg (commented):
I clicked skip mystery box button and Now I cannot see the box. I am in the allowlist with seven day streak
Screenshot 2024-12-13 at 7 07 30 PM

samott:
> I clicked skip mystery box button and Now I cannot see the box. I am in the allowlist with seven day streak

If you explicitly skip then you lose the box.

Yihwan:
> > I clicked skip mystery box button and Now I cannot see the box. I am in the allowlist with seven day streak
> 
> If you explicitly skip then you lose the box.

hi! if it's not too much trouble, for streaks can we have "skip" just close the modal? and then the user can come back to it simply by clicking the streaks stat card again (which would still display the Claim your CHOMPmas box state). 

lmk if this would be complicated; i'm not sure if this was explicitly mentioned anywhere, and could be confusing given how we were ok skipping the mystery box for the claim all flow.,

marvinmarnold:
> hi! if it's not too much trouble, for streaks can we have "skip" just close the modal? and then the user can come back to it simply by clicking the streaks stat card again (which would still display the Claim your CHOMPmas box state).
> 
> lmk if this would be complicated; i'm not sure if this was explicitly mentioned anywhere, and could be confusing given how we were ok skipping the mystery box for the claim all flow.,

How about we keep current skip behavior for this PR then update desired behavior in: https://github.com/gator-labs/chomp/pull/738 @neel-ds 

@Yihwan do we want different skip behavior for Claim v CHOMPmas? Or should skipping both now cause https://github.com/gator-labs/chomp/pull/738 to show up?

PR #742, opened 2024-12-16
PROD-569: Fix allowlist script ignore empty

neel-ds:
- Description
Filter out empty records from allowlist csv before creating records

Use add allowlist users script

NA

### DIFF
--- a/scripts/addAllowlistUsers.ts
+++ b/scripts/addAllowlistUsers.ts
@@ -31,8 +31,11 @@ async function main() {
   const csv = fs.readFileSync(csvFilePath, "utf8");
   const rows = csv.split(/\r?\n/);
 
+  // Filter out empty rows before processing
+  const validRows = rows.filter((address: string) => address.trim().length > 0);
+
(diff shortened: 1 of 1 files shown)


yash431garg (approved):
LGTM

scripts/addAllowlistUsers.ts
```diff
@@ -31,8 +31,11 @@ async function main() {
...
 
+  // Filter out empty rows before processing
+  const validRows = rows.filter((address: string) => address.trim().length > 0);
+
   await prisma.mysteryBoxAllowlist.createMany({
-    data: rows.map((address: string) => ({
+    data: validRows.map((address: string) => ({
       address: address.trim(),
       tags,
     })),
```
samott, line 41: Do the tags need any validation too? E.g. any restrictions on allowed characters (whitespace, ...)?

Not a blocker since this is a dev tool.

neel-ds: Not really, there is no as such restrictions on tags

PR #745, opened 2024-12-16
PROD-579/do-not-display-streak-info-after-claiming-streak

danijel-radomiljac:


marvinmarnold:
@danijel-radomiljac are you looking into this?

PR #760, opened 2024-12-17
fix: sentry and mixpanel claim, reveal flush

yash431garg:


marvinmarnold (approved):

neel-ds (approved):
Follow up on the events recorded in Mixpanel based on specific transaction outcomes.

PR #763, opened 2024-12-18
PROD-590/always-flush-sentry-if-used 

danijel-radomiljac:


app/actions/chompResult.ts
```diff
@@ -245,6 +247,7 @@ export async function revealQuestions(
         newFatl: revealPoints,
       },
     });
+    await Sentry.flush(SENTRY_FLUSH_WAIT);
```
marvinmarnold, line 250: The task description wasn't clear. But the intention is that we only call flush exactly once, just before the server action/endpoint returns.

marvinmarnold (changes requested):

marvinmarnold (approved):

marvinmarnold:
@danijel-radomiljac conflicts fyi

PR #764, opened 2024-12-18
PROD-596: change delete and create query to update query

yash431garg:


samott (approved):

neel-ds (approved):

app/actions/chompResult.ts
```diff
@@ -157,20 +157,7 @@ export async function revealQuestions(
   const revealPoints = await calculateRevealPoints(questionRewards);
 
   await prisma.$transaction(async (tx) => {
```
marvinmarnold, line 159: Previously, we were doing everything inside a DB transaction because we were temporarily deleting entries, then adding them back later. From looking at the code now, I think all of the DB actions we take are atomic. This means that we don't have to do everything in a single DB transaction. Wdyt @samott?

samott: Yeah, that makes sense... as long as there is no issue with the ChompResult state potentially not being consistent.

app/actions/chompResult.ts
```diff
@@ -192,20 +179,62 @@ export async function revealQuestions(
...
+              burnTransactionSignature: burnTx,
+              transactionStatus: TransactionStatus.Pending,
+            },
+            data: {
+              burnTransactionSignature: burnTx,
+              rewardTokenAmount: qr.rewardAmount,
+              transactionStatus: TransactionStatus.Completed,
+              revealNftId,
+            },
+          }),
```
samott, line 234: @yash431garg is the reward amount actually being updated here? If not, maybe this could be an `updateMany()`

yash431garg: Yea, reward amount is getting updated here.

app/actions/chompResult.ts
yash431garg: This is implemented by devin

samott (approved):

neel-ds (approved):

yash431garg:
@samott should we add this as well in the release?

samott:
> @samott should we add this as well in the release?

sure

PR #768, opened 2024-12-19
PROD-592: Fix reveal status mixpanel event

neel-ds:
- Description
This update ensures accurate event recording by marking it as RevealTransactionPending instead of RevealTransactionSucceeded when the transaction status cannot be confirmed in some edge cases (e.g delay in sign txn)
 
Try to reveal and sign txn after a few seconds of delay, you can verify from mixpanel dashboard in preview env

Screenshot 2024-12-19 at 6 17 59 PM


neel-ds:
I mistakenly created branch from my previous work. Please ignore the other changes related to mysteryBox and migration.

scripts/addAllowlistUsers.ts
```diff
@@ -31,8 +31,18 @@ async function main() {
   const csv = fs.readFileSync(csvFilePath, "utf8");
   const rows = csv.split(/\r?\n/);
 
-  // Filter out empty rows before processing
-  const validRows = rows.filter((address: string) => address.trim().length > 0);
+  // Filter out invalid addresses before adding to database
+  const validRows = rows.filter((address: string) => {
```
marvinmarnold, line 35: These changes are unrelated to the PR

prisma/migrations/20241218180555_address_length_check/migration.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "MysteryBoxAllowlist"
```
marvinmarnold, line 1: Unrelated to PR

app/hooks/useReveal.ts
```diff
@@ -349,16 +350,31 @@ export function useReveal({
...
+      const revealResult = await getRevealPending(revealQuestionIds);
+
+      if (revealResult?.length !== 0) {
+        trackEvent(TRACKING_EVENTS.REVEAL_TRANSACTION_PENDING, {
+          transactionSignature: signature,
+          nftAddress: revealNft?.id,
+          nftType: revealNft?.type,
+          burnedAmount: reveal?.amount,
+          [TRACKING_METADATA.REVEAL_TYPE]:
+            revealQuestionIds.length > 1 ? REVEAL_TYPE.ALL : REVEAL_TYPE.SINGLE,
```
marvinmarnold, line 362: It is possible to reveal all with only one question. It's not a high priority to fix this. How hard would it be to differentiate based on the button the user is clicking instead of the number of questions?

app/hooks/useReveal.ts
```diff
@@ -349,16 +350,31 @@ export function useReveal({
...
-        transactionSignature: signature,
-        nftAddress: revealNft?.id,
-        nftType: revealNft?.type,
-        burnedAmount: reveal?.amount,
-        [TRACKING_METADATA.REVEAL_TYPE]:
-          revealQuestionIds.length > 1 ? REVEAL_TYPE.ALL : REVEAL_TYPE.SINGLE,
-        [TRACKING_METADATA.QUESTION_ID]: reveal?.questionIds,
-        [TRACKING_METADATA.QUESTION_TEXT]: reveal?.questions,
-      });
+      const revealResult = await getRevealPending(revealQuestionIds);
```
marvinmarnold, line 353: revealResult not the best name. Maybe `pendingResults`?

app/actions/chompResult.ts
```diff
@@ -251,6 +251,22 @@ export async function revealQuestions(
   revalidatePath("/application");
 }
 
+export async function getRevealPending(questionIds: number[]) {
```
marvinmarnold, line 254: We should start putting all new actions into their own files. It's a preference, but it's a bit clearer to review the code that way.

marvinmarnold: Also this name could be improved, maybe: getPendingChompResults

yash431garg: We already have something for that https://github.com/gator-labs/chomp/blob/b9d8159639023e1149434985cbf5d1f14b0b3b93/app/actions/chompResult.ts#L341-L342

yash431garg (commented):
Try asking devin to fix.

app/hooks/useReveal.ts
neel-ds: Correct but our logic is bound to questions length so event will be recorded based on number of questions. Can you elaborate the intent?

app/actions/chompResult.ts
neel-ds: Thank you @yash431garg

app/hooks/useReveal.ts
neel-ds: I updated all REVEAL mixpanel events with a flag passed as an args from button. REVEAL_TYPE will be based on button and not questionIds.

neel-ds:
I updated all REVEAL mixpanel events with a flag passed as an args from button. REVEAL_TYPE will be based on button and not questionIds. We are GTG!

PR #773, opened 2024-12-20
PROD-597/prevent-reveal-and-reveal-all-if-no-or-low-bonk-balance

danijel-radomiljac:


yash431garg:
@danijel-radomiljac I think it still allowing user to reveal without bonk if user have nft and he do reveal with bonk.

samott (approved):

PR #777, opened 2024-12-20
PROD-578: Fix box single click 

neel-ds:
- Description
In merge conflicts, create mysteryBox based on allowlist logic was removed and this PR reverts it. Additionally, it enables single click on box for opening and adjust box size once opened.

You can open unclaimed mystery box prize on reveal tab and open it via box instead of button

NA


app/actions/claim.ts
```diff
@@ -91,9 +91,9 @@ export async function claimAllAvailable() {
 
   if (!claimableQuestionIds.length) throw new Error("No claimable questions");
 
-  const FF_MYSTERY_BOX = process.env.NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL;
+  const isUserAllowed = await isUserInAllowlist();
 
-  const mysteryBoxId = FF_MYSTERY_BOX
+  const mysteryBoxId = isUserAllowed
```
samott, line 96: I thought the allowlist was only for Chompmas?

neel-ds: The allowlist was not limited to Chompmas. Its purpose is to replace the feature flag and provide initial access to the mystery box for a selective group of users, which cannot be achieved through FF alone.

app/actions/claim.ts
```diff
@@ -91,9 +91,9 @@ export async function claimAllAvailable() {
 
   if (!claimableQuestionIds.length) throw new Error("No claimable questions");
 
-  const FF_MYSTERY_BOX = process.env.NEXT_PUBLIC_FF_MYSTERY_BOX_CLAIM_ALL;
+  const isUserAllowed = await isUserInAllowlist();
```
yash431garg, line 94: Can you rename it to something like isEligibleForMysteryBox?

neel-ds:
@yash431garg I am aware of it but in mobile screen it would compensate that space. It won't overlap with the above heading.

app/actions/claim.ts
neel-ds: yeah, will do it

neel-ds:
I made the box size responsive according to screen using tailwind screen clamps.

yash431garg (dismissed):
Look small in desktop but can be fixed in a follow up task.

neel-ds:
> Look small in desktop but can be fixed in a follow up task.

Hope recent commit would do the needful
Screenshot 2024-12-23 at 8 11 28 PM

yash431garg (approved):

samott (approved):

PR #781, opened 2024-12-23
PROD-608: Add valid burn sign constraint

neel-ds:
- Description
This PR will prevent empty string burn tx in database and record in Sentry. It adds a constraint and validate in server actions using bs58 decode if string is valid signature or not

NA

NA

### DIFF
--- a/app/actions/chompResult.ts
+++ b/app/actions/chompResult.ts
@@ -1,6 +1,10 @@
 "use server";
 
-import { RevealConfirmationError, RevealError } from "@/lib/error";
+import {
+  InvalidBurnTxError,
+  RevealConfirmationError,
+  RevealError,
+} from "@/lib/error";
 import {
   FungibleAsset,
   NftType,
@@ -11,7 +15,7 @@ import * as Sentry from "@sentry/nextjs";
 // import { TransactionSignature } from "@solana/web3.js";
 import { revalidatePath } from "next/cache";
 
-// import { SENTRY_FLUSH_WAIT } from "../constants/sentry";
+import { SENTRY_FLUSH_WAIT } from "../constants/sentry";
 import { questionAnswerCountQuery } from "../queries/questionAnswerCountQuery";
 import prisma from "../services/prisma";
 import { calculateCorrectAnswer, calculateReward } from "../utils/algo";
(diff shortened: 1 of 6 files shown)


app/actions/chompResult.ts
```diff
@@ -171,6 +175,18 @@ export async function revealQuestions(
     return null;
   }
 
+  if (!isValidSignature(burnTx)) {
```
yash431garg, line 178: Can you do this before we do hasBonkBurnedCorrectly?

prisma/migrations/20241223180534_burn_valid_signature/migration.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE "ChompResult" 
+ADD CONSTRAINT "burnTransactionSignature_not_empty" 
+CHECK ("burnTransactionSignature" IS NULL OR LENGTH("burnTransactionSignature") > 0);
```
marvinmarnold, line 3: Looking at ChompResults, burnTransactionSignature should always be between 86-88 characters.

neel-ds: Do we want to limit signature length between this range?

app/utils/solana.ts
```diff
@@ -132,3 +133,16 @@ export const getSolBalance = async (address: string): Promise<number> => {
 
   return balance / LAMPORTS_PER_SOL;
 };
+
+export function isValidSignature(
+  signature: string | null | undefined,
+): boolean {
+  if (!signature) return true; // Null is allowed but no empty string or random string
```
samott, line 140: doesn't this allow empty strings, contrary to the comment?

marvinmarnold: Screenshot 2024-12-23 at 11 08 24 AM

samott (dismissed):

neel-ds:
Gentle Reminder: We need to remove existing invalid burn tx records from prod before merging this.

app/utils/solana.ts
```diff
@@ -132,3 +133,19 @@ export const getSolBalance = async (address: string): Promise<number> => {
 
   return balance / LAMPORTS_PER_SOL;
 };
+
+export function isValidSignature(
+  signature: string | null | undefined,
+): boolean {
+  // Null is allowed but no empty string or random string
+  if (signature === null || signature === undefined) return true;
+  if (signature.length < 85 || signature.length > 89 || signature === "")
```
marvinmarnold, line 142: signature === "" is unnecessary now

PR #788, opened 2024-12-24
refactor: new query for total points and codebase refactoring for cre…

yash431garg:


yash431garg:
@marvinmarnold this PR also includes the fix for build failure.

PR #798, opened 2024-12-25
Prod 610: allow payout even no bonk

yash431garg:
- Description
Previously it wasn't possible to add bonk to a new wallet. 
Create a new wallet and try to claim bonk. It should be succesfull.
N/A


yash431garg:
@marvinmarnold it cost $0.4 to create user mint account. If it doesn't exist.

PR #809, opened 2024-12-27
PROD-610: add associated address if doesn't exists, optimize send bonk an…

yash431garg:
- Description
Previously it wasn't possible to add bonk to a new wallet. 
Create a new wallet and try to claim bonk. It should be succesfull.
N/A

### DIFF
--- a/app/utils/claim.ts
+++ b/app/utils/claim.ts
@@ -1,8 +1,8 @@
 "use server";
 
-import { ChompResult } from "@prisma/client";
 import * as Sentry from "@sentry/nextjs";
 import {
+  createAssociatedTokenAccountInstruction,
   createTransferInstruction,
   getAssociatedTokenAddress,
 } from "@solana/spl-token";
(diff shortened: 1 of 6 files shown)


yash431garg:
@marvinmarnold I have rebased this branch with main and squashed as well. If you think rebase is correct we can merge this work and close #798

app/utils/claim.ts
```diff
@@ -71,24 +67,40 @@ export const sendBonk = async (
 
   const bonkMint = new PublicKey(process.env.NEXT_PUBLIC_BONK_ADDRESS!);
 
-  const fromTokenAccount = await getAssociatedTokenAddress(
+  const treasuryATA = await getAssociatedTokenAddress(
```
neel-ds, line 70: Can you use more descriptive name like treasuryTokenAccount?

app/utils/claim.ts
```diff
@@ -71,24 +67,40 @@ export const sendBonk = async (
...
     bonkMint,
     fromWallet.publicKey,
   );
-  const toTokenAccount = await getAssociatedTokenAddress(bonkMint, toWallet);
 
-  const instruction = createTransferInstruction(
-    fromTokenAccount,
-    toTokenAccount,
+  const receiverATA = await getAssociatedTokenAddress(bonkMint, toWallet);
```
neel-ds, line 75: same goes here with name

app/utils/claim.ts
```diff
@@ -71,24 +67,40 @@ export const sendBonk = async (
...
+    );
+    instructions.push(ataInstruction);
+  }
+
+  const transferInstruction = createTransferInstruction(
+    treasuryATA,
+    receiverATA,
     fromWallet.publicKey,
     amount,
+    [],
```
neel-ds, line 97: Can you add a comment explaining the empty array argument?

app/utils/claim.ts
```diff
@@ -113,16 +125,16 @@ export const sendBonk = async (
   instructions.unshift(computeBudgetIx);
 
   //refered from past transaction
-  const computeUnitFix = 4794;
-  // Buffer to make sure the transaction doesn't fail because of less compute units
-  const computeUnitsIx = ComputeBudgetProgram.setComputeUnitLimit({
-    units: Math.round(computeUnitFix * 1.1),
-  });
+  // const computeUnitFix = 4794;
```
neel-ds, line 128: is it fine to remove CU optimization here if it is not required?

app/utils/claim.ts
yash431garg: It is not a token account it is AssociatedTokenAddress which means it is the mint address for that account. If you want I can make it treasuryAssociatedTokenAddress.

app/utils/claim.ts
yash431garg: Sure,

app/utils/claim.ts
yash431garg: Yea, I have about will ask in the call regarding this.

marvinmarnold (approved):

neel-ds (approved):

PR #811, opened 2024-12-27
PROD-620: Deck Screen CTA

neel-ds:
- Description
Show the buttons according to credits required and add buy credit drawer

Open a deck

Screenshot 2024-12-27 at 10 43 34 PM
Screenshot 2024-12-27 at 10 43 47 PM

### DIFF
--- a/app/components/DeckScreenAction/DeckScreenAction.tsx
+++ b/app/components/DeckScreenAction/DeckScreenAction.tsx
@@ -3,7 +3,7 @@ import { TRACKING_EVENTS, TRACKING_METADATA } from "@/app/constants/tracking";
 import trackEvent from "@/lib/trackEvent";
 import { CloseIcon } from "@dynamic-labs/sdk-react-core";
 import { DialogTitle } from "@radix-ui/react-dialog";
-import { CircleArrowRight } from "lucide-react";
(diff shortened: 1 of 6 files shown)


app/components/DeckScreenAction/DeckScreenAction.tsx
```diff
@@ -0,0 +1,107 @@
...
+  currentDeckId: number;
+  setIsDeckStarted: (isDeckStarted: boolean) => void;
+  totalCredits: number;
+  deckCost: number;
+};
+
+const CREDIT_COST_FEATURE_FLAG =
+  process.env.NEXT_PUBLIC_FF_CREDIT_COST_PER_QUESTION === "true";
+
+const DeckScreenAction = ({
```
yash431garg, line 23: I think we can just keep the name DeckScreen as it in a separate folder or we can keep the name deckscreencomponent. What do you think?

yash431garg: Otherwise LGTM... can be merged.

neel-ds: We should avoid using the same component name that already exists. `DeckScreenAction` feels more appropriate as “DeckScreenCTA” for clarity. What do you think?

marvinmarnold (approved):

PR #814, opened 2024-12-30
Prod 594: adjust deck stack new

yash431garg:
- Description
Updated stacks code to show first the expiring date based on whichever is expiring first comes first, than show upcoming decks based on whichever is upcoming first based on activeFromDate and then show expired decks based on whichever expired first comes first.
Decks must be sorted based on the above comment.
N/A

### DIFF
--- a/app/queries/stack.ts
+++ b/app/queries/stack.ts
@@ -39,10 +39,10 @@ export async function getStacks() {
 
 export async function getStack(id: number) {
   const jwt = await getJwtPayload();
-
   const userId = jwt?.sub;
+  const now = new Date();
 
-  return prisma.stack.findUnique({
+  const stack = await prisma.stack.findUnique({
     where: {
       id,
     },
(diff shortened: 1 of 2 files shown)


neel-ds (approved):

app/queries/stack.ts
```diff
@@ -74,6 +74,66 @@ export async function getStack(id: number) {
       },
     },
   });
+
+  if (!stack) {
+    return null;
+  }
+
+  const sortedDecks = [...stack.deck].sort((a, b) => {
+    if (
```
marvinmarnold, line 83: This won't work for the daily deck stack since they don't have activeFromDate

yash431garg: Yea, we have separate stacks for daily deck. That's why I didn't added that. Should that also need to be handled.
Screenshot 2024-12-30 at 9 27 05 PM

marvinmarnold: After looking at daily deck stack, it looks like it's already in a reasonable order. I think you are right that we don't need to handle it now.

Screenshot 2024-12-30 at 11 56 03 AM

PR #817, opened 2024-12-31
fix: credit deck cost condition

neel-ds:
- Description
Fix the DeckScreen CTA action conditions


yash431garg (approved):
I think comments can be removed as keywords are self explanatory

PR #819, opened 2025-01-01
PROD-630: show box animation based on box type

yash431garg:
- Description
Update logic based on box type instead of prize for the box to show proper animation.
Closed green box for new user after user completes or skip the tutorial.
Screenshot 2025-01-01 at 12 04 22 PM

### DIFF
--- a/components/MysteryBox/MysteryBox.tsx
+++ b/components/MysteryBox/MysteryBox.tsx
@@ -271,7 +271,7 @@ function MysteryBox({
               animationData={
                 boxType === EMysteryBoxType.Chompmas
                   ? animationDataSanta
-                  : EMysteryBoxType.Tutorial
+                  : boxType === EMysteryBoxType.Tutorial
                     ? animationDataCredits
                     : animationDataRegular
               }
--- a/components/MysteryBox/MysteryBox.tsx
+++ b/components/MysteryBox/MysteryBox.tsx
@@ -271,7 +271,7 @@ function MysteryBox({
               animationData={
                 boxType === EMysteryBoxType.Chompmas
                   ? animationDataSanta
-                  : creditsReceived > 0
+                  : EMysteryBoxType.Tutorial
                     ? animationDataCredits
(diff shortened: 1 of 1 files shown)


neel-ds (approved):

components/MysteryBox/MysteryBox.tsx
```diff
@@ -271,7 +271,7 @@ function MysteryBox({
               animationData={
                 boxType === EMysteryBoxType.Chompmas
                   ? animationDataSanta
-                  : creditsReceived > 0
+                  : EMysteryBoxType.Tutorial
```
marvinmarnold, line 274: `EMysteryBoxType.Tutorial` won't this always evaluate to true? I think it needs to be `boxType === EMysteryBoxType.Tutorial`.

But also based on [this](https://linear.app/gator/issue/PROD-591/mechanism-engine-should-return-correct-minimums), the claim/reveal all flow may give credits or bonk or both. I don't think this logic considers that.

marvinmarnold (changes requested):

components/MysteryBox/MysteryBox.tsx
yash431garg: Good catch, yea it needs to be boxType === EMysteryBoxType.Tutorial. Yea, this logic doesn't consider the case if user get both the rewards. Actually this task is only related to show green box for tutorial mystery box because currently we were showing blue closed box in the start. I think we have a seperate task to handle the animation when we have both the reward. https://linear.app/gator/issue/PROD-627/award-credits-for-chompmas-and-claim-all-mystery-boxes

marvinmarnold:
FYI I lost track of this similar issue. I added it to this cycle. It's about making sure the box can only be awarded once. https://linear.app/gator/issue/PROD-633/tutorial-complete-mystery-box-polish

yash431garg:
Tested for all three boxes type working fine.

marvinmarnold (approved):

PR #825, opened 2025-01-02
feat: make send bonk utils server only

yash431garg:
Using server only make sure that if this is imported in client it will through error.

### DIFF
--- a/app/utils/claim.ts
+++ b/app/utils/claim.ts
@@ -1,5 +1,3 @@
-"use server";
-
 import { ChompResult } from "@prisma/client";
 import * as Sentry from "@sentry/nextjs";
 import {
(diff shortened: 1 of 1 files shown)


yash431garg:
@marvinmarnold It is not tested trying to check if we can use this as it is recommended in the nextjs docs.

yash431garg:
Also @samott already using this in mystery box https://github.com/gator-labs/chomp/blob/81a8e771c32b0d4d73e7011120e6a7531f9ced0d/lib/mysteryBox.ts#L17

cc: @marvinmarnold

marvinmarnold (approved):
I found [this thread](https://www.reddit.com/r/nextjs/comments/18ih6nd/useserver_vs_serveronly/?rdt=59931) helpful, even the very last message.

My understanding is that the core question is: do you want clients to be able to directly invoke the code?
- Yes: "use server"
- No: "server-only"

In this specific case, it looks like we never invoke `app/utils/claim.ts#sendBonk` directly from client, only via other server side functions.

PR #826, opened 2025-01-03
Prod 582: charge user credit per question

yash431garg:
- Description
User will be charged for paid decks on per question basis and record will stored in fatl table. This PR also restricts chomping a question if user reaches insufficient in the middle of deck. 
1. Create paid decks
2. Credits should reduce after a paid deck is strated or completed.
3. A Drawer will pop up if user balance reaches low in the middle 
4. Deck will start from begin if user return to home and start again. 
N/A

### DIFF
--- a/app/actions/fungible-asset.ts
+++ b/app/actions/fungible-asset.ts
@@ -68,6 +68,38 @@ export const incrementFungibleAssetBalance = async ({
   return result[0];
 };
 
+// charge user credits for premium decks/questions
+export const chargeUserCredits = async (questionId: number) => {
+  const payload = await authGuard();
+
+  const getCredits = await prisma.deck.findMany({
+    where: {
+      deckQuestions: {
+        some: {
+          questionId,
+        },
+      },
+    },
+    select: {
+      creditCostPerQuestion: true,
+    },
+  });
+
+  const creditCostPerQuestion = getCredits[0]?.creditCostPerQuestion;
+
+  if (creditCostPerQuestion === null) return;
+
+  await prisma.fungibleAssetTransactionLog.create({
+    data: {
+      type: TransactionLogType.PremiumQuestionCharge,
+      questionId: questionId,
+      asset: FungibleAsset.Credit,
+      change: -Math.abs(creditCostPerQuestion),
+      userId: payload.sub,
+    },
+  });
+};
+
 export const addXPoints = async () => {
   const payload = await authGuard();
 
--- a/app/actions/fungible-asset.ts
+++ b/app/actions/fungible-asset.ts
@@ -1,6 +1,8 @@
 "use server";
 
+import { QuestionMultiDecksError } from "@/lib/error";
 import { FungibleAsset, TransactionLogType } from "@prisma/client";
+import * as Sentry from "@sentry/nextjs";
 import { revalidatePath } from "next/cache";
 
 import { pointsPerAction } from "../constants/points";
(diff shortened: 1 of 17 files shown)


yash431garg:
Migration is required for credit deduction.

app/actions/fungible-asset.ts
```diff
@@ -68,6 +68,38 @@ export const incrementFungibleAssetBalance = async ({
   return result[0];
 };
 
+// charge user credits for premium decks/questions
+export const chargeUserCredits = async (questionId: number) => {
+  const payload = await authGuard();
+
+  const getCredits = await prisma.deck.findMany({
```
marvinmarnold, line 75: Confusing variable name makes it sound like a function. Something like `creditsForQuestions` would be clearer.

app/actions/fungible-asset.ts
```diff
@@ -68,6 +68,38 @@ export const incrementFungibleAssetBalance = async ({
...
+          questionId,
+        },
+      },
+    },
+    select: {
+      creditCostPerQuestion: true,
+    },
+  });
+
+  const creditCostPerQuestion = getCredits[0]?.creditCostPerQuestion;
```
marvinmarnold, line 88: I think the assumption is that every question should belong to exactly one deck. So we can just look at the 0th index. In that case, would be better to explicitly throw an error if more than one deck returned. Should trigger Sentry event and user should still be able to reveal other questions.

If we have a DB constraint which ensures each question will only ever belong to one deck, then this code is fine. Otherwise we should explicitly check.

app/queries/deck.ts
```diff
@@ -135,6 +135,11 @@ export async function getDeckQuestionsForAnswerById(deckId: number) {
 
   const totalDeckQuestions = getTotalNumberOfDeckQuestions(deckQuestions);
 
+  const deckCost =
+    deck?.creditCostPerQuestion !== null
+      ? deck?.creditCostPerQuestion * deckQuestions.length
```
marvinmarnold, line 140: Would be safer to `sum(questions.creditCost) in deck` than to assume all questions will have same value. That is the assumption now, but might as well make the code more robust now.

app/queries/deck.ts
```diff
@@ -135,6 +135,11 @@ export async function getDeckQuestionsForAnswerById(deckId: number) {
 
   const totalDeckQuestions = getTotalNumberOfDeckQuestions(deckQuestions);
 
+  const deckCost =
```
marvinmarnold, line 138: If this variable is only used for credits I would use `deckCreditCost`, otherwise I think it could be used for BONK too.

app/components/Deck/Deck.tsx
```diff
@@ -61,6 +63,8 @@ type DeckProps = {
   deckId: number;
   nextDeckId?: number;
   deckVariant?: "daily-deck" | "regular-deck";
+  deckCost: number | null;
+  CREDIT_COST_FEATURE_FLAG: boolean;
```
marvinmarnold, line 67: All caps here doesn't make sense here. From `Deck`'s perspective, this is a regular variable with snakeCase.

marvinmarnold (changes requested):

app/queries/deck.ts
yash431garg: Right now the keyword creditCostPerQuestion is associated with the deck and not with every question that's why I have to write it in the this way.

PR #823, opened 2025-01-01
PROD-631: Move Reward Mystery Box to Reveal All

neel-ds:
- Description
Reward Mystery Box is moved to Reveal All success. It adds one more TriggerType.

- What are the steps to test that this code is working?
Try to reveal all from history page and you will get mystery box instead of after claim all

- Screen shots or recordings for UI changes
NA

PR #827, opened 2025-01-03
PROD-627: Enable credits with bonk mystery box

neel-ds:
- Description
This will create mysterybox with credits and bonk if credit is present in reward and credit animation will be visible after revealAll event type.

Reveal all and open mystery box to verify 

NA

### DIFF
--- a/lib/mysteryBox.ts
+++ b/lib/mysteryBox.ts
@@ -11,6 +11,7 @@ import {
   EBoxPrizeType,
   EBoxTriggerType,
   EMysteryBoxStatus,
+  EPrizeSize,
 } from "@prisma/client";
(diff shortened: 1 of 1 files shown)


neel-ds:
Note: Tests are failing because this is forked from #823 which requires migration

marvinmarnold:
@neel-ds can you create and merge a PR into demo<-main, after the migration has been solved on main: https://linear.app/gator/issue/PROD-631/move-existing-mystery-box-flow-that-appears-after-claimshare-to-after#comment-1f8bcf66

yash431garg (commented):
Code LGTM... How can I test the pr? I mean which box have both the rewards?

PR #832, opened 2025-01-06
fix: mystery box reveal all logic

neel-ds:
- Description
Remove allowlist validation from regular REVEAL_ALL mystery box and only enabled for History page via optional args

NA


.env.example
```diff
@@ -42,6 +42,7 @@ FF_CHOMP_BOT=
 
 # Mystery Box Feature Flag
 NEXT_PUBLIC_FF_MYSTERY_BOX_CHOMPMAS=
+NEXT_PUBLIC_FF_MYSTERY_BOX=
```
marvinmarnold, line 45: Name this NEXT_PUBLIC_FF_MYSTERY_BOX_REVEAL_ALL

marvinmarnold (changes requested):

.env.example
neel-ds: Done

__tests__/actions/mystery-box.test.ts
```diff
@@ -221,7 +221,6 @@ describe("Create mystery box", () => {
       expect(res?.id).toBe(mysteryBoxId);
       expect(res?.status).toBe("New");
       expect(res?.triggers).toHaveLength(1);
-      expect(res?.MysteryBoxPrize).toHaveLength(1);
```
marvinmarnold, line 224: Ideally, we wouldn't decrease the specificity of the test. Eg this could be `expect(prize) >= 1` (pseudocode) instead of deleting it all together.

app/actions/chompResult.ts
```diff
@@ -258,16 +259,15 @@ export async function revealQuestions(
...
-      if (isEligibleForMysteryBox) {
-        await rewardMysteryBox(
-          payload?.sub,
-          EBoxTriggerType.RevealAllCompleted,
-          revealableQuestionIds,
-        );
-      }
+    if (
+      isMysteryBoxEnabled &&
+      process.env.NEXT_PUBLIC_FF_MYSTERY_BOX === "true"
```
marvinmarnold, line 264: Need to update env var everywhere: NEXT_PUBLIC_FF_MYSTERY_BOX_REVEAL_ALL

marvinmarnold (changes requested):
Don't forget to set NEXT_PUBLIC_FF_MYSTERY_BOX_REVEAL_ALL correctly in Vercel also.

__tests__/actions/mystery-box.test.ts
neel-ds: Gotcha

app/actions/chompResult.ts
neel-ds: Oh! I made the change but file was unsaved due to format error.

PR #844, opened 2025-01-08
Prod 656: fake sol payment

yash431garg:
- Description
1. Create a chain tx when user initiate credit purchase
2. Update FATL for the chainTx
3. Refresh user state
It require migration before testing
N/A


app/components/BuyCreditsDrawer/BuyCreditsDrawer.tsx
```diff
@@ -8,9 +11,38 @@ import { Drawer, DrawerContent } from "../ui/drawer";
...
+
+  const [isProcessingTx, setIsProcessingTx] = useState(false);
+
+  const { promiseToast } = useToast();
+
+  const router = useRouter();
+
+  const processTx = async () => {
+    setIsProcessingTx(true);
+    await promiseToast(createBuyCreidtsTx(deckId), {
```
neel-ds: Spelling mistake in credits

app/actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,96 @@
...
+
+/**
+ * Create a new chainTx - Save the entry in chain tx table when
+ * user initiate credit purchase, create sol tx and call update fatl action.
+ *
+ * @param deckId The backend should validate the amount of credit
+ *               purchased based on deck id
+ */
+
+export async function createBuyCreidtsTx(deckId: number) {
```
neel-ds, line 20: Spelling mistake in credits

app/actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,96 @@
...
+  const payload = await getJwtPayload();
+
+  if (!payload) return;
+
+  const creditForQuestion = await prisma.deck.findMany({
```
neel-ds, line 25: It should be `findUnique` instead of findMany as it queries single deck.

neel-ds (changes requested):

prisma/schema.prisma
```diff
@@ -295,6 +297,22 @@ model MysteryBoxAllowlist {
   MysteryBoxTrigger MysteryBoxTrigger[]
 }
 
+model ChainTx {
+  hash                        String                        @id @default(uuid())
```
marvinmarnold, line 301: Let's not make this the default at the DB level. Shouldn't be possible to accidentally create without a hash. Instead, manually generate a UUID for now and supply it with the create command.

eslint.config.mjs
```diff
@@ -33,6 +33,7 @@ const config = [
       "@typescript-eslint/no-explicit-any": "off",
       "@typescript-eslint/no-non-null-asserted-optional-chain": "off",
       "@typescript-eslint/no-empty-object-type": "off",
+      "prettier/prettier": 0,
```
marvinmarnold, line 36: What does this do?

app/components/BuyCreditsDrawer/BuyCreditsDrawer.tsx
```diff
@@ -8,9 +18,56 @@ import { Drawer, DrawerContent } from "../ui/drawer";
...
+  const totalSolCost = Number(solPricePerCredit) * creditsToBuy;
+
+  const [isProcessingTx, setIsProcessingTx] = useState(false);
+
+  const router = useRouter();
+
+  const processTx = async () => {
+    setIsProcessingTx(true);
+
+    await createBuyCreidtsTx(deckId)
```
marvinmarnold, line 41: typo

marvinmarnold (changes requested):

eslint.config.mjs
yash431garg: There was a annoying issue for indentation because of conflict in eslint and prettier. It solves that.

app/actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,96 @@
...
+  });
+
+  if (!wallet) {
+    throw new Error(
+      "Unable to prepare transaction. Don't worry, nothing was submitted on-chain. Please try again",
+    );
+  }
+
+  const fromWallet = Keypair.fromSecretKey(
+    base58.decode(process.env.CHOMP_TREASURY_PRIVATE_KEY || ""),
```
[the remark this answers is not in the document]
yash431garg: If we import something like this CHOMP_TREASURY_PRIVATE_KEY in a client component it will be undefined. If we need to import something like this in a client component we need to make it like NEXT_PUBLIC_CHOMP_TREASURY_PRIVATE_KEY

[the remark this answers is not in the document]
yash431garg: I think it a default property of nextjs

app/actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,96 @@
...
+    !creditForQuestion[0].creditCostPerQuestion ||
+    creditForQuestion[0].deckQuestions.length < 1 ||
+    process.env.NEXT_PUBLIC_SOLANA_COST_PER_CREDIT === null
+  ) {
+    throw new Error(
+      "Unable to prepare transaction. Don't worry, nothing was submitted on-chain. Please try again",
+    );
+  }
+
+  const deckCost =
```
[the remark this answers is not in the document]
yash431garg: Make sense, I wasn't aware that we have this column with every question as well.

eslint.config.mjs
yash431garg: Yes, It deactivate the error for prettier. I updated the code to take it as a warning instead of error but CI check will still work as it is and format it properly.

marvinmarnold (approved):

neel-ds (approved):

PR #846, opened 2025-01-08
Restore streak for 1st Jan

neel-ds:
Add query to find affected users and solution to add activity manually for affected users


streak.md
```diff
@@ -0,0 +1,92 @@
...
+JOIN jan2_users j2 ON u.id = j2."userId"
+LEFT JOIN jan1_users j1 ON u.id = j1."userId"
+WHERE j1."userId" IS NULL;
+```
+
+### Restore Streak of affected users
+
+```sql
+-- Insert QuestionAnswers
+INSERT INTO "QuestionAnswer" ("questionOptionId", "userId", "percentage", "status", "selected", "timeToAnswer", "createdAt", "updatedAt")
```
marvinmarnold, line 48: It looks like this has to be run manually for each user and each question? Ideally, one command could be used to update for all impacted users

streak.md
```diff
@@ -0,0 +1,92 @@
...
+SELECT 
+  6610,
+  "userId",
+  50,
+  'Submitted'::text::public."AnswerStatus",
+  random() < 0.5,
```
marvinmarnold, line 54: Won't this mean that the user may have more than one 'selected' answer for the same question?

streak.md
```diff
@@ -0,0 +1,92 @@
...
+  6000,
+  TIMESTAMP '2024-01-01 12:00:00.000',
+  TIMESTAMP '2024-01-01 12:00:00.000'
+FROM (VALUES ('USER_1'), ('USER_2')) as t("userId")
```
marvinmarnold, line 58: I don't understand what's going on here. Is USER_1 and USER_2 supposed to be replaced with user IDs?

streak.md
```diff
@@ -0,0 +1,92 @@
...
+
+-- Insert ChompResults
+INSERT INTO "ChompResult" ("userId", "questionId", "result", "transactionStatus", "rewardTokenAmount", "burnTransactionSignature", "createdAt", "updatedAt")
+SELECT 
+  "userId",
+  2859,
+  'Revealed'::text::public."ResultType",
+  'Completed'::text::public."TransactionStatus",
+  0,
+  '5KtPn3MZJGkXGYx6K7BbNNa9Cy3eLSmop1SXfUJJQYPhzyqbYKhqp95GNzBqafEJQGwqWBk4Brh8Dv9PupZYPXPR',
```
marvinmarnold, line 79: What does this correspond to? Should use a more obvious fake value like 'STREAK_RECOVERY_NOT_REAL_TX'

marvinmarnold (changes requested):
I was expecting that the script could be run once and fix everything. This approach looks manual and error prone.

streak.md
neel-ds: Yes, we can include random() in the first query along with userId, and it can be used interchangeably in both scenarios.

streak.md
neel-ds: Correct but this can be avoided if we use nested CTE and get rid of manual values

streak.md
neel-ds: I recently added constraint in burnSignature that its length should be 86-88. This is fake signature.

marvinmarnold: Gotcha, then can we use: `STREAK_RECOVERY_NOT_REAL_TX_01_XXXXXXXXXXXXXX`. Fill with X until 86 characters long.

streak.md
marvinmarnold: Is it just the editor or the actual DB? If it's just the editor, you should be able to connect to Neon from DBeaver, or any other client, and do a CTE.

streak.md
marvinmarnold: I'm not clear what you are saying. We should make sure that each user has exactly one QuestionAnswer where selected = true.

neel-ds:
> @neel-ds query ran for me but there were few issues with the updated data. Can you implement the changes based on our discussion before implementing the script?

It was timestamp issue. I tested it. It works fine now. GTG!

marvinmarnold (approved):
lgtm. @neel-ds please run this manually then send an update in #product on slack letting them know.

Ideally, you could provide a spreadsheet of all the wallet addresses/usernames that were impacted.

neel-ds:
Request Processed.

PR #847, opened 2025-01-08
PROD-598: Save transaction signatures before burn

samott:
This accomplishes two things for reveals:

- We save transaction signatures after signing but before transmitting to the blockchain.
- If an exception happens, don't delete chomp results.


app/hooks/useReveal.ts
```diff
@@ -407,10 +406,6 @@ export function useReveal({
         setRevealNft(undefined);
       }
     } catch (error) {
-      if (pendingChompResultIds.length > 0) {
```
marvinmarnold, line 410: @yash431garg what do you think about this change?

app/hooks/useReveal.ts
```diff
@@ -332,6 +329,8 @@ export function useReveal({
           );
 
           pendingChompResultIds = chompResults?.map((cr) => cr.id) ?? [];
+
+          await CONNECTION.sendRawTransaction(tx.serialize());
```
marvinmarnold, line 333: It would probably be a little more robust to send the tx server side, at the end of createQuestionChompResults.

marvinmarnold (commented):
Will test it now

app/hooks/useReveal.ts
samott: Good idea, I hadn't considered that.

app/hooks/useReveal.ts
samott: I definitely welcome thoughts about this - are there any consequences for having the hanging results? I also considered adding a new enum status value instead of deleting them but the end result is perhaps the same.

We always have a burn signature when we create these results so any other exceptions must be after the transaction was created.

marvinmarnold: The hanging results will prevent the user from being able to reveal again. In at least some cases, there will be a burn signature but the tx doesn't land on-chain because fees were too low.

But I defer to @yash431garg as he's been dealing with this code more.

marvinmarnold (approved):

app/hooks/useReveal.ts
yash431garg: I think it fine to remove this code as it is a complicated piece of code and we also have seen few entries which are succeed but deleted from chompResult and we can keep the hanging results because our cron job will handle that eventually and we already have a new column needManualReview which will also be marked by cron job based on tx status.

app/actions/chompResult.ts
```diff
@@ -385,6 +386,28 @@ export async function createQuestionChompResult(
   return results;
 }
 
+export async function createChompResultsAndSubmitSignedTx(
```
yash431garg, line 389: Can you add comments for this actions? Can you add sentry here to catch the error or a new mixpanel track event for the error?

yash431garg (changes requested):
I have added few comments.

marvinmarnold (approved):

yash431garg (approved):

PR #854, opened 2025-01-09
feat: add cron job for low treasury balance

yash431garg:
- Description
Removed treasury alert from claim action and created a cron to make sure that we aren't sending sentry error for every user. 
Treasury alert api should send a alert if invoked
N/A


app/api/treasury-balancer-alert/route.ts
marvinmarnold, line 1: nit: should be treasury-balance-alert (no r). Balancer makes it sound like it's moving funds around.

marvinmarnold (approved):

app/api/treasury-balancer-alert/route.ts
yash431garg: Sorry, It was a typo

neel-ds (approved):

marvinmarnold (approved):

PR #856, opened 2025-01-11
PROD-658: Purchase Credits with SOL 

neel-ds:
- Description
This PR allows one to buy credits with SOL and update Chain Tx. It fixes small issue too. 
Screenshot 2025-01-11 at 1 49 46 AM

https://solana.fm/tx/3rmBV4ojY3LEZLF5HNbkn7t4xE9drdXngfHx6w3bX2rcGAsGxf393N5acbgqXZPQhLYLi14DAr7Q955fkLpnHW5R?cluster=mainnet-alpha

Open high credit-based deck and buy some credits

https://github.com/user-attachments/assets/2e1a1fd8-7573-42ad-829f-ac5296f515ce


app/actions/credits/createChainTx.ts
```diff
@@ -1,48 +1,38 @@
...
 
-export async function createBuyCreditsTx(creditToBuy: number) {
+export async function createSignedSignatureChainTx(
+  creditsToBuy: number,
+  signature: string,
+) {
   const payload = await getJwtPayload();
 
-  const treasuryKey = getTreasuryPrivateKey();
+  if (!payload) throw new Error("User not authenticated");
```
yash431garg, line 27: I have configured the errors in a way that it will catch the exact error message in the frontend. Can you configure it according to that?

lib/credits/processTransaction.ts
```diff
@@ -0,0 +1,107 @@
...
+  // Send transaction
+  const txHash = await CONNECTION.sendRawTransaction(
+    signedTransaction.serialize(),
+    { skipPreflight: true },
+  );
+
+  try {
+    // Wait for confirmation
+    await pRetry(async () => {
+      const currentBlockhash = await CONNECTION.getLatestBlockhash();
```
yash431garg, line 48: Out of scope for this ticket.

lib/credits/processTransaction.ts
```diff
@@ -0,0 +1,107 @@
...
+    });
+    await Sentry.flush(SENTRY_FLUSH_WAIT);
+    throw new Error("Transaction failed to confirm");
+  }
+
+  try {
+    // Wait for finalization
+    await pRetry(async () => {
+      const latestBlockhash = await CONNECTION.getLatestBlockhash();
+      await CONNECTION.confirmTransaction(
```
yash431garg, line 79: Similarly Out of scope.

lib/credits/initiatePurchase.ts
```diff
@@ -0,0 +1,70 @@
...
+  creditsToBuy: number,
+  wallet: Wallet,
+  setIsProcessingTx: (isProcessingTx: boolean) => void,
+) {
+  const payload = await getJwtPayload();
+  if (!payload) {
+    throw new Error("User not authenticated");
+  }
+
+  const release = await acquireMutex({
```
yash431garg, line 32: Out of scope

app/actions/credits/createChainTx.ts
```diff
@@ -54,21 +44,19 @@ export async function createBuyCreditsTx(creditToBuy: number) {
...
+  }
 
   try {
-    newChainTx = await prisma.chainTx.create({
+    await prisma.chainTx.create({
       data: {
-        hash: uuidv4(),
+        hash: signature,
         status: EChainTxStatus.New,
         solAmount: String(solAmount),
```
yash431garg, line 61: I think this data needs to be populated in the db. 

Correctly populate feeSolAmount for the ChainTx. It may not be possible to know the exact fee amount at this time

If that's the case, then add a new column, estimatedFeeSolAmount and store the value there

yash431garg (changes requested):

app/actions/credits/updateChainTxStatus.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Update the chain transaction status to finalized and create Credit FATL
+ *
+ * @param hash - The hash of the transaction
+ * @param creditAmount - The amount of credits purchased
+ * @throws Error if transaction fails to update
+ */
+export async function updateTxStatusToFinalized(
```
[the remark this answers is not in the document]
marvinmarnold: Also from what I can tell, this method is only ever called from other server actions. If that's the case, this does not need to be a server action, it can be a regular function.

lib/credits/processTransaction.ts
```diff
@@ -0,0 +1,122 @@
...
+        { cause: error },
+      );
+    Sentry.captureException(transactionFailedToFinalizeError, {
+      extra: {
+        creditAmount: creditsToBuy,
+        signature: txHash,
+      },
+    });
+    await Sentry.flush(SENTRY_FLUSH_WAIT);
+    throw new Error("Transaction failed to finalize");
```
marvinmarnold, line 120: This is a bit confusing. I would expect all trys to either throw an exception or captureException to Sentry. If we do both, we risk reporting the error to sentry twice. I think you can just throw TransactionFailedToFinalizeError and that will result in the error being eventually logged to Sentry.

lib/priorityFee.ts
```diff
@@ -0,0 +1,63 @@
...
+  // Buffer to make sure the transaction doesn't fail because of less compute units
+  const modifyComputeUnits = ComputeBudgetProgram.setComputeUnitLimit({
+    units: Math.round(computeUnitFix * 1.1),
+  });
+  tx.add(modifyComputeUnits);
+
+  let estimateFee = await getRecentPrioritizationFees(tx);
+
+  // Verify the estimateFee is not null due to RPC request failure in some cases
+  if (estimateFee === null) {
```
marvinmarnold, line 37: p-retry?

marvinmarnold (changes requested):
Not all of these comments are critical, I generally defer to you about what's important.

neel-ds:
> Not all of these comments are critical, I generally defer to you about what's important.
> 
> Also, why are the tests failing? Is it related to this? https://linear.app/gator/issue/PROD-669/flaky-deck-test

Yes it's part of flaky test issue, if I re-run failed one then it will work fine.

lib/constant.ts
```diff
@@ -0,0 +1,8 @@
+// CONSTANT VALUES
+
+export const getTreasuryPublicKey = () => {
```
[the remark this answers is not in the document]
neel-ds: I had this idea, but using server actions in client-side components might expose them. I ensured it doesn’t involve the private key and only uses the public key if required. Let me know your thoughts, I appreciate your input.

app/actions/credits/createChainTx.ts
```diff
@@ -55,20 +52,22 @@ export async function createBuyCreditsTx(creditToBuy: number) {
...
         wallet: wallet?.address,
```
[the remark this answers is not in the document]
neel-ds: Should I update the schema to add an id field for the Wallet table, or keep using the address as the default id? Let me know.

app/actions/credits/updateChainTxStatus.ts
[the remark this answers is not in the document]
neel-ds: make sense i will move it to lib

app/actions/credits/updateChainTxStatus.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  const payload = await getJwtPayload();
+
+  if (!payload) {
+    return {
+      error: "User not authenticated",
+    };
+  }
+
+  while (attempt < MAX_RETRIES && !success) {
```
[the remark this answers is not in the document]
neel-ds: This seems to be from Yash’s PR for simulating SOL transactions. I’ll update it to use p-retry.

lib/constant.ts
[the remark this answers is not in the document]
marvinmarnold: I don't think there's a risk of exposing an env var that doesn't have NEXT_PUBLIC and is in a file with "use server".

.env.example
```diff
@@ -26,6 +26,7 @@ MIX_PANEL_TOKEN=
 # Treasuery
 CHOMP_TREASURY_PRIVATE_KEY=
 CHOMP_SQUADS=
+NEXT_PUBLIC_TREASURY_PUBLIC_ADDRESS=
```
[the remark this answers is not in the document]
marvinmarnold: CHoMP5YdLEJ62kq9oibKbNDkBCgakQPqQLSgkDHyC2D9

app/actions/credits/createChainTx.ts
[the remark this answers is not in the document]
marvinmarnold: Ah ok I see now that Wallet table doesn't have an independent ID. The primary key is the address. So it's actually fine the way it is now.

app/components/PreviewDeckCard/PreviewDeckCard.tsx
```diff
@@ -78,7 +78,7 @@ const PreviewDeckCard = ({
           >
             <span className="opacity-50 pr-1">Entry </span>
             {deckCreditCost > 0
-              ? `${deckCreditCost} Credit${deckCreditCost !== 1 && "s"}`
+              ? `${deckCreditCost} Credit${deckCreditCost !== 1 ? "s" : ""}`
```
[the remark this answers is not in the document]
neel-ds: This refactoring solves this issue (mentioned in PR description)

Screenshot 2025-01-11 at 1 49 46 AM

lib/priorityFee.ts
neel-ds: sure

lib/credits/processTransaction.ts
```diff
@@ -0,0 +1,122 @@
...
+          ...latestBlockhash,
+        },
+        "finalized",
+      );
+
+      const txInfo = await CONNECTION.getTransaction(txHash, {
+        maxSupportedTransactionVersion: 0,
+      });
+
+      const fees = txInfo?.meta?.fee;
```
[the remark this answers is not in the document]
neel-ds: yes total fees spent on transaction

lib/credits/initiatePurchase.ts
```diff
@@ -0,0 +1,95 @@
...
+) {
+  const payload = await getJwtPayload();
+  if (!payload) {
+    return {
+      error: "User not authenticated",
+    };
+  }
+
+  const release = await acquireMutex({
+    identifier: "CREDIT_PURCHASE",
```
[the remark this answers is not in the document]
neel-ds: I saw use of mutex in [PROD-657](https://linear.app/gator/issue/PROD-657/verify-payment-details). Shall I remove it from here?

lib/credits/initiatePurchase.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    if (!data) {
+      return {
+        error: "Transaction declined",
+      };
+    }
+
+    const { transaction, signature } = data;
+
+    // Step 2: Record the signed transaction in ChainTx
```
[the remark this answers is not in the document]
neel-ds: Step 3: ProcessTransaction requires Transaction w/ signature and it cannot be passed as prop to server from client function. I tried Shaun's approach to substitute the supported prop type but it didn't work. I kept it as it is for now until we find any potential solution.

actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,97 @@
...
+) {
+  const payload = await getJwtPayload();
+
+  if (!payload) {
+    return {
+      error: "User not authenticated",
+    };
+  }
+
+  const SOLANA_COST_PER_CREDIT = process.env.NEXT_PUBLIC_SOLANA_COST_PER_CREDIT;
```
marvinmarnold, line 34: Since this is in a function scope, it shouldn't be all caps. `solanaCostPerCredit`

actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  if (!SOLANA_COST_PER_CREDIT) {
+    return {
```
marvinmarnold, line 37: Returning {error} here is not consistent with how we typically throw Errors when something goes wrong. It also doesn't seem to follow any of the conventions: https://nextjs.org/docs/app/building-your-application/routing/error-handling

That said, I'm not sure it's worth the time to update this.

actions/credits/createChainTx.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  try {
+    await prisma.chainTx.create({
+      data: {
+        hash: signature,
+        status: EChainTxStatus.New,
+        solAmount: String(solAmount),
+        wallet: wallet?.address,
+        feeSolAmount: "0",
```
marvinmarnold, line 74: Is this being set to 0 because feeSolAmount is a required field?

lib/credits/processTransaction.ts
neel-ds: I ensured that it doesn't record errors twice in Sentry by removing captureException from initiatePurchase() because it handles errors in two ways: 

1. Errors that we directly show to user via toast (e.g Transaction declined, Unable to prepare transaction, please try again) are being handled by `return { error: "ERROR_MESSAGE" }`
2. Errors (related to transaction / confirmation) are not disclosed directly so it handles this via try-catch (throw) and at the end user will see Transaction Failed, ...contact support

actions/credits/createTransaction.ts
marvinmarnold, line 1: This isn't a server action without "use server". That makes sense because all this logic happens on the client. This file should be moved to lib or somewhere else.

lib/credits/initiatePurchase.ts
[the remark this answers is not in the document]
marvinmarnold: During the verification step, credits are awarded if successful. In that context, a mutex is appropriate to avoid credits being awarded multiple times for a single payment.

In this case, I think it's unnecessary. But it doesn't hurt much to keep if you want.

lib/credits/updateTxStatusConfirm.ts
```diff
@@ -0,0 +1,47 @@
+"use server";
```
marvinmarnold, line 1: "use server" not needed unless you want to invoke this directly from a client as a server action.

lib/credits/updateTxStatusFinalized.ts
```diff
@@ -0,0 +1,70 @@
+"use server";
```
marvinmarnold, line 1: Remove

lib/credits/updateTxStatusFinalized.ts
```diff
@@ -0,0 +1,70 @@
...
+          await tx.chainTx.update({
+            where: { hash },
+            data: {
+              status: EChainTxStatus.Finalized,
+              finalizedAt: new Date(),
+              feeSolAmount: feesInSOL ? feesInSOL.toString() : undefined,
+            },
+          });
+
+          await tx.fungibleAssetTransactionLog.create({
```
marvinmarnold, line 47: Ah OK, I see now that processTransaction calls this ultimately. This part justifies the mutex I incorrectly said wasn't necessary.

lib/credits/initiatePurchase.ts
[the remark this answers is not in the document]
marvinmarnold: After further looking, you are right that the mutex can be useful here. This code leads to points being given out.

lib/credits/updateTxStatusFinalized.ts
```diff
@@ -0,0 +1,70 @@
...
+
+  if (!payload) throw new Error("Unauthorized access");
+
+  const userId = payload.sub;
+
+  try {
+    await pRetry(
+      async () => {
+        await prisma.$transaction(async (tx) => {
+          await tx.chainTx.update({
```
marvinmarnold, line 38: I'm curious what happens if processTransaction is called twice with the same transaction. Is it possible that it will incorrectly allow users to process the same transaction multiple times and get credits each time?

lib/priorityFee.ts
```diff
@@ -0,0 +1,63 @@
...
+) {
+  const { blockhash, lastValidBlockHeight } =
+    await CONNECTION.getLatestBlockhash("confirmed");
+
+  tx.recentBlockhash = blockhash;
+  tx.lastValidBlockHeight = lastValidBlockHeight;
+  tx.feePayer = feePayer;
+
+  // It is recommended to add the compute limit instruction before adding other instructions
+  const computeUnitFix = 4960;
```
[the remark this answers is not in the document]
marvinmarnold: Can you add the link to comments in the code?

Also, that transaction only seems to use 1,874 units. 

Screenshot 2025-01-15 at 2 55 45 PM

marvinmarnold (changes requested):
I forgot to say last time, but this is missing test cases.

.env.example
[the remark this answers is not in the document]
marvinmarnold: I think you can remove this now, since public key is now derived from the private key.

lib/priorityFee.ts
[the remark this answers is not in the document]
neel-ds: Yes sure, I think we can reduce it 2k around.

actions/credits/createChainTx.ts
neel-ds: yes it's required field

samott:
I'm seeing this warning:

```
Warning: Only plain objects can be passed to Client Components from Server Components. Decimal objects are not supported.
  {id: 19, userId: ..., questionId: 1, burnTransactionSignature: ..., sendTransactionSignature: ..., rewardTokenAmount: Decimal, result: ..., createdAt: ..., updatedAt: ..., transactionStatus: ..., needsManualReview: ..., revealNftId: ...}
```

I think Prisma leverages Decimal.js when it is available -- so rewardTokenAmount should be converted to a string or number.

lib/credits/updateTxStatusFinalized.ts
neel-ds: Added FATL unique constraint

actions/credits/createChainTx.ts
marvinmarnold: Not blocking comments:
- Do we currently separate estimated fees from actual fees? Maybe we should.
- We should at least know the estimated fees at this point. I'd rather store those estimates and then overwrite with actual on confirmation.

marvinmarnold (changes requested):
Unit tests are a blocker for me. New actions and things in lib dir.

neel-ds:
@marvinmarnold I’ve added tests for all actions. However, I didn’t include tests for creating Solana transactions, signing with the signer from Dynamic, or processing transactions. This is because these only support fresh transaction confirmations, whereas a different method is ideal for constant transactions in tests.

For estimated fees, I tried using [getFeeForMessage()](https://www.quicknode.com/docs/solana/getFeeForMessage) to fetch recent confirmed slot fees, but it returned null for our compiled transaction (even signed tx). Using priority fee for estimation doesn’t seem reasonable. Since we already get the exact fee (feesSOL) once the transaction is finalized, the estimated fee property might not be necessary. Let me know if there is different approach that you are thinking off

This PR also aligns with the new file structure — keeping route files in the app directory while others remain outside.

app/components/BuyCreditsDrawer/BuyCreditsDrawer.tsx
```diff
@@ -28,27 +29,30 @@ function BuyCreditsDrawer({
 }: BuyCreditsDrawerProps) {
   const solPricePerCredit = process.env.NEXT_PUBLIC_SOLANA_COST_PER_CREDIT;
   const totalSolCost = Number(solPricePerCredit) * creditsToBuy;
```
samott, line 31: We should avoid doing arithmetic on currency values with floats - better to use decimal.js for this, even if it's converted back to a number after.

app/components/BuyCreditsDrawer/BuyCreditsDrawer.tsx
```diff
@@ -65,51 +69,56 @@ function BuyCreditsDrawer({
...
           isLoading={isProcessingTx}
-        >
-          Buy Credits
-        </Button>
-        <Button onClick={onClose} variant="outline">
-          Close
-        </Button>
-      </DrawerContent>
-    </Drawer>
+          loadingMessage="Processing transaction..."
```
samott, line 79: The text is not centered relative to Chompy.

marvinmarnold:
> Using priority fee for estimation doesn’t seem reasonable. Since we already get the exact fee (feesSOL) once the transaction is finalized, the estimated fee property might not be necessary. Let me know if there is different approach that you are thinking off

I'm mixing two different comments so it's confusing.

1) Making sure the DB never contains invalid data. I consider a temporary 0 fee as invalid data because we know that's not the case.
2) Knowing how well we are estimating fees. For this, we could add an extra column.

But I think only comment #1 should potentially be in scope for this PR. I guess I'm ultimately suggesting that fee is made a nullable field.

yash431garg (approved):
LGTM

actions/credits/verifyPayment.ts
```diff
@@ -0,0 +1,136 @@
...
+        // Protects against scenario where money flows to the treasury in
+        // one instruction but then back out in another (could potentially
+        // still be valid if the net amount is correct, but there are
+        // many edge cases to consider if we allow this...)
+        if (
+          parsed.type === "transfer" &&
+          parsed.info.source === treasuryWallet &&
+          parsed.info.lamports > 0
+        ) {
+          return false;
```
marvinmarnold, line 96: For future validations, it would be ideal to return context about why it failed validation. That way error messages can be more clear.

marvinmarnold (approved):
Thanks for reorganizing things @neel-ds. Personally, I find the code a lot easier to understand and navigate now. I tested and working.

samott (approved):
Works great. Nice job!

PR #860, opened 2025-01-15
Prod 494: Support subj mcq

yash431garg:
Removed the refinement rule that required at least one correct option in multi-choice questions.


yash431garg:
@marvinmarnold the code already handles the populating calculated percentage for multiple options and mech-engine also updates the same endpoint. That's why this pr doesn't have any functional changes.

cc: @mleiwe

marvinmarnold (approved):

samott (approved):

PR #864, opened 2025-01-16
fix: change link tag to anchor tag for revalidation

yash431garg:
- Description
Stacks weren't showing the updated data because the data wasn't revalidated.

### DIFF
--- a/app/components/StackCard/StackCard.tsx
+++ b/app/components/StackCard/StackCard.tsx
@@ -2,7 +2,6 @@
 
 import { STACKS_PATH } from "@/lib/urls";
 import Image from "next/image";
-import Link from "next/link";
 
 import { ArrowRightCircle } from "../Icons/ArrowRightCircle";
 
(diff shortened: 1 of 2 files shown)


marvinmarnold:
Why doesn't prefetch work with our code?

yash431garg:
> Why doesn't prefetch work with our code?

My explanation was wrong previously. It wasn't a prefetch issue. The issue which this ticket resolve is that the stack page wasn't revalidating after user completes a deck. One option we can have is to use revalidate but the issue is the that we have to add that in deck action which is common for both home and decks in stacks and revalidation will reload the page to get the new data. That's why I just replaced the Link tag to anchor because by default anchor tag will reload after user complete the deck and we will get new data in stacks page. Also, I find out that we use anchor tag for the decks in application page to make sure that the data is revalidated https://github.com/gator-labs/chomp/blob/3881c48cb7039119f72f94a00b2008de6fd5016f/app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx#L49 That's why I used this approach.

neel-ds (approved):

marvinmarnold (approved):

samott (approved):

PR #865, opened 2025-01-16
PROD-664 added check for wallet address in rank items for all time leaderboard

AndrejRadomiljac:
Link to the Linear task 
https://linear.app/gator/issue/PROD-664/400-error-on-all-time-leaderboard-on-stagingchompgames
- Description
I've managed to reproduce it on staging via wallet address that Yihwan mentioned in the task, unfortunately (luckily) not on PROD. I think something is related to accounts (wallets) so I've added conditional checks that will prevent breaking the page.
- After the fix
Screenshot 2025-01-13 at 16 45 35

### DIFF
--- a/app/components/LeaderboardRanking/LeaderboardRanking.tsx
+++ b/app/components/LeaderboardRanking/LeaderboardRanking.tsx
@@ -30,8 +30,8 @@ const LeaderboardRanking = ({ label, loggedUserId, ranking }: Props) => {
           {ranking.map((rankItem) => {
             const name = rankItem.user.username
               ? `@${rankItem.user.username}`
-              : rankItem.user.wallets[0].address
-                ? formatAddress(rankItem.user.wallets[0].address)
+              : rankItem.user.wallets[0]?.address
+                ? formatAddress(rankItem.user.wallets[0]?.address)
                 : "mocked user";
             return (
               <RankingCard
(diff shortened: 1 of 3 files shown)


samott (approved):

app/components/LeaderboardRanking/LeaderboardRanking.tsx
```diff
@@ -30,8 +30,8 @@ const LeaderboardRanking = ({ label, loggedUserId, ranking }: Props) => {
           {ranking.map((rankItem) => {
             const name = rankItem.user.username
               ? `@${rankItem.user.username}`
-              : rankItem.user.wallets[0].address
-                ? formatAddress(rankItem.user.wallets[0].address)
+              : rankItem.user.wallets[0]?.address
+                ? formatAddress(rankItem.user.wallets[0]?.address)
                 : "mocked user";
```
marvinmarnold, line 35: I know 'mocked user' was already here, but it seems like it shouldn't. Can you create a follow up ticket to remove this?

marvinmarnold (commented):
I'd like us to have add tests everytime we fix bugs. In this case, you could refactor name calculation into a helper method that you unit test. 

These nested ternary ifs are pretty confusing to read and should generally be broken up.

AndrejRadomiljac:
> I'd like us to have add tests everytime we fix bugs. In this case, you could refactor name calculation into a helper method that you unit test.
> 
> These nested ternary ifs are pretty confusing to read and should generally be broken up.

I agree with you, but since we do not have test component (UI) setup I was not able to do that.

samott (approved):

PR #868, opened 2025-01-17
fix: increase confirmation timeout

yash431garg:


samott (commented):
Looks good, although it seems like we could be using pRetry in this code?

yash431garg:
> Looks good, although it seems like we could be using pRetry in this code?

Can we do it in a follow task instead?

samott:
> > Looks good, although it seems like we could be using pRetry in this code?
> 
> Can we do it in a follow task instead?

Yes, ok, since we need to get this wrapped up.

samott (approved):

marvinmarnold (approved):

PR #869, opened 2025-01-20
PROD-684: Fix slow query for unauthenticated sessions.

samott:
Update query to not fetch all chomp results in the database when we are querying in an unauthenticated session.

Prevents large amounts of data being sent to the frontend.

### DIFF
--- a/__tests__/queries/stack/get-stack-results.test.ts
+++ b/__tests__/queries/stack/get-stack-results.test.ts
@@ -5,10 +5,7 @@ import prisma from "@/app/services/prisma";
 import { authGuard } from "@/app/utils/auth";
 import { faker } from "@faker-js/faker";
 import {
-  AnswerStatus,
-  QuestionType,
   ResultType,
-  Token,
   TransactionStatus,
 } from "@prisma/client";
 import dayjs from "dayjs";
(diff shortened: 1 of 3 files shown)


yash431garg:
@samott I think this pr doesn't resolve: 
The current query fetches all the question answer and chompResults for that deck causing data overload and also causing the error which is passing obj from server to client for chompResults

samott:
> @samott I think this pr doesn't resolve: The current query fetches all the question answer and chompResults for that deck causing data overload and also causing the error which is passing obj from server to client for chompResults

I think this does solve the main issue as only the user's results are fetched when logged in. However, I have some more optimisations coming later today. Will keep you updated.

marvinmarnold (commented):
Please add test

yash431garg (approved):
LG

yash431garg:
A quick question out of curiosity. Have you tested it against replica that you aren't get 404 for drip deck?

samott:
> A quick question out of curiosity. Have you tested it against replica that you aren't get 404 for drip deck?

Yes, I tested against stack 42. No errors with with the PR applied.

neel-ds (approved):

PR #874, opened 2025-01-20
PROD-659: Process Pending Credit Payment [Cron]

neel-ds:
### Description
- API checks New Chain Tx records from the past week that are not finalized and not failed.
- It validates these transactions by:
  - If finalized:
	•	Updates the Chain Tx status to Finalized
	•	Deposits credits to the user’s account.
  - If not finalized:
	•	Records a Sentry event.
	•	Updates the status to failedAt with the current timestamp.


app/api/cron/process-credits/route.ts
```diff
@@ -0,0 +1,114 @@
...
+    });
+
+    for (const tx of pendingTransactions) {
+      const isValid = await verifyCreditPayment(tx.hash);
+
+      if (isValid) {
+        const walletOwner = await getWalletOwner(tx.wallet);
+
+        if (!walletOwner) {
+          throw new Error("Wallet owner not found");
```
marvinmarnold, line 55: This will cause the whole loop to fail, but other entries should be allowed to continue fine. We should also update failedAt for the failed tx.

app/api/cron/process-credits/route.ts
```diff
@@ -0,0 +1,114 @@
...
+ * This API processes pending credit purchase transactions that fall within
+ * the 1 week to 10 minutes before the current time.
+ *
+ * Actions performed:
+ * 1. Attempts to validate the payment transaction hash
+ * 2. If successful, marks the chainTx as `finalized` and adds credits to user
+ * 3. If unsuccessful, logs error to Sentry
+ */
+
+export async function GET(request: Request) {
```
marvinmarnold, line 24: This could be broken up into some smaller methods for readibility.

app/api/cron/process-credits/route.ts
```diff
@@ -0,0 +1,114 @@
...
+
+      if (isValid) {
+        const walletOwner = await getWalletOwner(tx.wallet);
+
+        if (!walletOwner) {
+          throw new Error("Wallet owner not found");
+        }
+
+        // Update chainTx status and add credits
+        await prisma.$transaction(async (prisma) => {
```
marvinmarnold, line 59: Feels like it should be shared logic with the main tx verification flow.

app/api/cron/process-credits/route.ts
```diff
@@ -0,0 +1,114 @@
...
+            data: {
+              userId: walletOwner,
+              change: creditAmount,
+              type: TransactionLogType.CreditPurchase,
+              asset: FungibleAsset.Credit,
+              chainTxHash: tx.hash,
+            },
+          });
+        });
+      } else {
```
marvinmarnold, line 83: Do we ever expect verifyCreditPayment to return false, then true if run later? Maybe if it's run immediately after a tx is submitted onchain?

Maybe every time this endpoint executes, it should check for txs that are older than 5 min but without failedAt, and fail those if still invalid. 10min is an arbitrary time corresponding to the longest I would expect it to take for a tx's state to settle.

lib/credits/verifyCreditPayment.ts
marvinmarnold, line 1: This file also looks like duplicate code from main verification flow.

app/api/cron/process-credits/route.ts
```diff
@@ -0,0 +1,114 @@
...
+              status: EChainTxStatus.New,
+            },
+            data: {
+              status: EChainTxStatus.Finalized,
+              finalizedAt: new Date(),
+            },
+          });
+
+          // Add credits based on solAmount
+          const creditAmount = Number(tx.solAmount) * 1000; // Credits per SOL
```
samott, line 72: Could this conversion rate by defined somewhere as a const or config value?

app/api/cron/process-credits/route.ts
neel-ds: Make sense added logic to capture in sentry and continue for rest

app/api/cron/process-credits/route.ts
neel-ds: I moved the verification of instructions into new function.

lib/credits/verifyCreditPayment.ts
neel-ds: Correct, this file use similar code but without payload guard and payload.sub checks

marvinmarnold (approved):

samott (approved):

PR #879, opened 2025-01-23
Prod 555/new home page without banner

yash431garg:
- Description
Updated deck section with feature flag
Screenshot 2025-01-23 at 3 18 02 PM


app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
```diff
@@ -44,7 +54,10 @@ export function HomeFeedDeckCard({
   status,
   date,
   deckId,
+  deckCreditCost,
 }: HomeFeedDeckCardProps) {
+  const CREDIT_COST_FEATURE_FLAG =
+    process.env.NEXT_PUBLIC_FF_CREDIT_COST_PER_QUESTION === "true";
```
samott, line 60: The tests elsewhere in the codebase only check if this flag is truthy. We should use a consistent convention.

samott (commented):
Overall, looks good and seems to work.

app/components/HomeFeedDeckCard/HomeFeedDeckCard.tsx
yash431garg: Can you elaborate more?

samott: elsewhere, we just do
`const CREDIT_COST_FEATURE_FLAG = process.env.NEXT_PUBLIC_FF_CREDIT_COST_PER_QUESTION;`
whereas you are requiring the flag to to be `== "true"`.

Either way is fine, but the flag won't work consistently across the codebase as it stands.

yash431garg: Make sense... I have seen some bugs without using this  == "true".

marvinmarnold (commented):
@yash431garg when I try to click into these questions, I'm not actually able to answer.

https://github.com/user-attachments/assets/e2b00c82-31d8-4939-9ed6-25bb6563779f

yash431garg:
> > @yash431garg when I try to click into these questions, I'm not actually able to answer.
> > Screen.Recording.2025-01-27.at.09.58.13.mov
> 
> You can try creating a new deck it is because of some wrong configuration in the deck data and which is also possible in previous expiring soon tab. I have also added a new deck Deck 1 you can try that and it is not possible in prod until we do some put some bad data like here.

marvinmarnold:
> > @yash431garg when I try to click into these questions, I'm not actually able to answer.
> > Screen.Recording.2025-01-27.at.09.58.13.mov
> 
> You can try creating a new deck it is because of some wrong configuration in the deck data and which is also possible in previous expiring soon tab. I have also added a new deck Deck 1 you can try that.

Can you clarify what the bad configuration is? Why are we confident that this issue won't happen when we release to production? It will also confuse the testers in staging.

yash431garg:
> > > @yash431garg when I try to click into these questions, I'm not actually able to answer.
> > > Screen.Recording.2025-01-27.at.09.58.13.mov
> > 
> > 
> > You can try creating a new deck it is because of some wrong configuration in the deck data and which is also possible in previous expiring soon tab. I have also added a new deck Deck 1 you can try that.
> 
> Can you clarify what the bad configuration is? Why are we confident that this issue won't happen when we release to production? It will also confuse the testers in staging.

I will come up with that after some more digging. I am not able to recall that right now.

yash431garg:
> > > > @yash431garg when I try to click into these questions, I'm not actually able to answer.
> > > > Screen.Recording.2025-01-27.at.09.58.13.mov
> > > 
> > > 
> > > You can try creating a new deck it is because of some wrong configuration in the deck data and which is also possible in previous expiring soon tab. I have also added a new deck Deck 1 you can try that.
> > 
> > 
> > Can you clarify what the bad configuration is? Why are we confident that this issue won't happen when we release to production? It will also confuse the testers in staging.
> 
> I will come up with the configuration to back this up after some more digging. I am not able to recall that right now.

yash431garg:
@marvinmarnold here is the reason behind that issue
Screenshot 2025-01-28 at 1 40 08 PM
Screenshot 2025-01-28 at 2 03 41 PM
Screenshot 2025-01-28 at 2 08 22 PM
Screenshot 2025-01-28 at 2 08 27 PM

PR #881, opened 2025-01-23
build(PROD-702): bump Dynamic for Fox Wallet support

marvinmarnold:
Description: Bump Dynamic to v4 in order to enable Fox Wallet support.

The auth token changes are not currently well implemented. See comments.


app/components/Bot/Bot.tsx
```diff
@@ -22,7 +23,8 @@ export default function Bot({
 }: {
   telegramAuthData: TelegramAuthDataProps;
 }) {
-  const { sdkHasLoaded, user, primaryWallet, authToken } = useDynamicContext();
+  const { sdkHasLoaded, user, primaryWallet } = useDynamicContext();
+  const authToken = getAuthToken();
```
marvinmarnold, line 27: Auth token use to be part of a hook but now it looks like a regular function. This means that it won't automatically update if it changes. It could even return null temporarily if it's called too early, before Dynamic context is loaded.

We should refresh the token whenever the dynamic context changes.

PR #883, opened 2025-01-24
PROD-688: manual banned users

samott:
Adds the ability to ban suspected bots.


lib/jwt.ts
```diff
@@ -0,0 +1,31 @@
...
+      await revokeDynamicSession(userId);
+    } catch (e) {
+      const dynamicRevokeSessionError = new DynamicRevokeSessionError(
+        `Failed revoke session for user with threat level: ${userId}`,
+        { cause: e },
+      );
+      Sentry.captureException(dynamicRevokeSessionError);
+      await Sentry.flush(SENTRY_FLUSH_WAIT);
+    }
+    throw new UserThreatLevelDetected("User threat level detected");
```
marvinmarnold, line 25: Could add userId and threadLevel to error message.

lib/jwt.ts
```diff
@@ -0,0 +1,31 @@
...
+} from "@/lib/error";
+import * as Sentry from "@sentry/nextjs";
+import { cookies } from "next/headers";
+
+export const checkThreatLevel = async (userId: string) => {
+  const user = await prisma.user.findFirst({ where: { id: userId } });
+  if (!user) throw new Error("User not found");
+  if (user.threatLevel) {
+    try {
+      await revokeDynamicSession(userId);
```
marvinmarnold, line 16: The way you've implemented this, I don't think revokeDynamicSession is even necessary. This implementation checks our DB for bot status on every request and also sends a revocation to Dynamic if threatLevel is present. But we never read from Dynamic's blocklist. So we may as well omit revokeDynamicSession all together.

The approach I had in mind was to eliminate all of decodeJwtPayload
Screenshot 2025-01-26 at 14 15 29 and replace it with a call to [this](https://docs.dynamic.xyz/api-reference/sdk/check-that-session-is-valid) Dynamic endpoint.

That said, your current approach is probably more efficient. Let's keep what you have and just remove revokeDynamicSession, since we aren't ever reading from that.

marvinmarnold (dismissed):

lib/jwt.ts
marvinmarnold: Oh I see. Ok, makes sense to omit.

marvinmarnold (approved):
You will need to manually upload the bots after merge, right?

samott:
> You will need to manually upload the bots after merge, right?

Yep

PR #888, opened 2025-01-27
PROD-669: Fix Home Deck Test

neel-ds:
- Description
Update the revealAtDate to current and changed the interval of years to past


samott (commented):
This still fails for me. I think the approach of changing the dates is causing problems over time that eventually break the tests again.

PR #891, opened 2025-01-27
Prod 682 author image fix

AndrejRadomiljac:

- Description
- Link to the task https://linear.app/gator/issue/PROD-682/add-deck-author-to-deck-info-page. 
- Admin should be able to add author name in "Deck Author"section while creating or editing deck.
- Admin should be able to add author image in "Deck Author Image" section while creating or editing deck.
- Both filed are optional
- Validation for image is set to the same "Default image size" that we have through whole app
- Deck author validation is set to minimum 5 chars. We can add upper limit later
- You'll see on many places border colour update, I've updated it to the colours that should be there. But the update was required due to Avatar image default border colour that was overwriting any colour from the outside.
- Migrations for deck table with two new columns
End results:
- Deck preview state with author image and author name
Screenshot 2025-01-23 at 16 45 48
- Deck preview state with author name, but no author image
Screenshot 2025-01-23 at 16 46 26
- Deck preview state with author image, but no name
Screenshot 2025-01-23 at 16 48 53
- Deck preview state with no author image and no name
Screenshot 2025-01-23 at 16 50 33

For not yet active decks we do not show author data
Screenshot 2025-01-23 at 16 51 18

### DIFF
--- a/app/schemas/deck.ts
+++ b/app/schemas/deck.ts
@@ -27,6 +27,21 @@ export const deckSchema = z
         }
         return true;
       }, "Only .jpg, .jpeg, .png and .webp formats are supported."),
+    authorImageFile: z
+      .custom<File[]>()
+      .optional()
+      .refine((files) => {
+        if (files && files.length > 0) {
+          return files[0].size <= IMAGE_UPLOAD_SIZES.DEFAULT;
+        }
+        return true;
+      }, `Max image size allowed is ${IMAGE_UPLOAD_SIZE_STRINGS.DEFAULT}.`)
+      .refine((files) => {
+        if (files && files.length > 0) {
+          return IMAGE_VALID_TYPES.includes(files[0].type);
+        }
+        return true;
+      }, "Only .jpg, .jpeg, .png and .webp formats are supported."),
     imageUrl: z
       .string()
       .optional()
@@ -47,6 +62,26 @@ export const deckSchema = z
         },
       ),
     description: z.string().min(5).optional().nullish().or(z.literal("")),
+    author: z.string().min(5).optional().nullish().or(z.literal("")),
+    authorImageUrl: z
+      .string()
+      .optional()
+      .nullable()
+      .refine(
+        (value) => {
+          if (!value) return true;
+
+          try {
+            new URL(value);
+            return true;
+          } catch {
+            return false;
+          }
+        },
+        {
+          message: "Invalid image source",
+        },
+      ),
     footer: z.string().min(5).max(50).optional().nullish().or(z.literal("")),
     tagIds: z.number().array().default([]),
     stackId: z.number().optional().nullish(),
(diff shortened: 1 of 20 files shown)


AndrejRadomiljac:
Here is the link to preview https://chomp-p8qbs1zfn-gator-labs.vercel.app/login?next=%2Fapplication%3FvercelToolbarCode%3Dmyij3T06BdyowH2. Add your account as admin, create deck using author image and name, make sure that your deck is active and you are good to go.

samott (approved):

app/schemas/deck.ts
```diff
@@ -47,6 +62,26 @@ export const deckSchema = z
         },
       ),
     description: z.string().min(5).optional().nullish().or(z.literal("")),
+    author: z.string().min(5).optional().nullish().or(z.literal("")),
```
marvinmarnold, line 65: Can we change the min to 2 chars?

marvinmarnold (changes requested):
The preview image looks a lot smaller for me than your screenshots

https://chomp-dev-git-prod-682-author-image-fix-gator-labs.vercel.app/application/decks/809

Screenshot 2025-01-27 at 10 16 13

marvinmarnold (changes requested):
Incorrect error message. Question description is in fact too short. But author name is optional. But the error message makes it seem like author name is required. Although the error message is misleading, actual behavior is correct. author name is actually treated as optional.
Screenshot 2025-01-27 at 10 17 23

app/schemas/deck.ts
AndrejRadomiljac: Sure 👍

AndrejRadomiljac:
> The preview image looks a lot smaller for me than your screenshots
> 
> https://chomp-dev-git-prod-682-author-image-fix-gator-labs.vercel.app/application/decks/809
> 
> Screenshot 2025-01-27 at 10 16 13

I'll check but it could be due image (effect). It should be 24x24

samott (approved):

AndrejRadomiljac:
@marvinmarnold I've fixed validation comments. Author name is now 2 chars instead of 5. Messages are fixed on all places including description, heading etc. One thing I've added, so if you type one char for author name and fill all other required parts, on submit will block deck creation since you've not achieved valid char limit even if it is optional. If you remove one char and leave that spot empty you will be able to submit successfully.

Regarding image size. Your image size is in proper dimension 24x24. The only thing that is tricky here is purple background and combination of colours of your image.
Screenshot 2025-01-27 at 22 08 32

marvinmarnold:
> @marvinmarnold I've fixed validation comments. Author name is now 2 chars instead of 5. Messages are fixed on all places including description, heading etc. One thing I've added, so if you type one char for author name and fill all other required parts, on submit will block deck creation since you've not achieved valid char limit even if it is optional. If you remove one char and leave that spot empty you will be able to submit successfully.

Great

> Regarding image size. Your image size is in proper dimension 24x24. 

Oh, turns out I was confused because I thought the deck image in your screeshots was the author image. So I was wondering why your images were so much bigger. But I was just looking at the wrong thing.

PR #894, opened 2025-01-28
PROD-714: Mystery box hub basic history

samott:
This implements the basic mystery box history page.

The history feed component is not actually installed anywhere as of yet. But I've attached a screenshot of the stand-alone component.

The small green mystery box image was not available as an SVG in the design, so I've thrown in the PNG file for now.

mystery_box_history


app/actions/mysteryBox/fetchHistory.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (
+      triggerType == "RevealAllCompleted" ||
+      triggerType == "DailyDeckCompleted" ||
+      triggerType == "ClaimAllCompleted"
+    )
+      category = EMysteryBoxCategory.Validation;
+    else if (triggerType == "TutorialCompleted")
+      category = EMysteryBoxCategory.Practice;
+    else category = EMysteryBoxCategory.Campaign;
```
marvinmarnold, line 85: I hadn't thought about how existing mystery boxes like ClaimAllCompleted would be handled. Good job supporting those.

Given that support, is it easy to add campaign boxes in this PR? Chompmas is the only one I think. We won't add campaigns to box opening flow. Just here, in the history view.

If it's harder than I realize, we don't have to.

marvinmarnold (dismissed):
Based on PR comment, I'm going to approve without testing.

app/actions/mysteryBox/fetchHistory.ts
samott: I think we can just switch off the filter in the query further up to enable the campaign boxes. The UI will show them.

marvinmarnold:
@yash431garg @neel-ds please review

components/MysteryBoxHub/MysteryBoxHistory.tsx
```diff
@@ -0,0 +1,62 @@
...
+          <div className="flex items-center justify-center py-8">
+            No mystery boxes!
+          </div>
+        )}
+
+        {mysteryBoxes?.map((box) => (
+          <MysteryBoxHistoryCard box={box} key={box.id} />
+        ))}
+
+        <LoadMoreDecks
```
yash431garg, line 52: Can you rename this to load more for clarity?

yash431garg (approved):

app/actions/mysteryBox/fetchHistory.ts
```diff
@@ -0,0 +1,97 @@
...
+  const records = await prisma.mysteryBox.findMany({
+    where: {
+      userId: payload.sub,
+      status: EMysteryBoxStatus.Opened,
+      triggers: {
+        some: {
+          triggerType: {
+            in: [
+              EBoxTriggerType.ClaimAllCompleted,
+              EBoxTriggerType.ClaimAllCompleted,
```
neel-ds, line 31: I think it's added twice. It is meant to be DailyDeckCompleted, right?

neel-ds (approved):

neel-ds (approved):

app/actions/mysteryBox/fetchHistory.ts
samott: Yes, good catch. I've removed the filter per @marvinmarnold's comment.

yash431garg (approved):

yash431garg:
LGTM

neel-ds (approved):

PR #896, opened 2025-01-29
Prod 691/enable hide deck new

yash431garg:
- Description
1. Add hideDeckFromHome flag in stacks
2. Hide deck from home if the flag is turn on. 
Only deck with hideDeck from home with value false or null. Should be visible.

### DIFF
--- a/app/components/FreeDeckFeed/FreeDeckFeed.tsx
+++ b/app/components/FreeDeckFeed/FreeDeckFeed.tsx
@@ -29,7 +29,11 @@ function FreeDeckFeed() {
       return [...acc, ...page];
     }, []);
   }, [data]);
-  if (formattedData?.length === 0 && !isFetching) {
+
+  if (
+    (formattedData?.length === 0 || formattedData === undefined) &&
+    !isFetching
+  ) {
     return <NoDeck />;
   }
   return (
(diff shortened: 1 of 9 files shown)


samott (approved):
Looks good.

app/components/FreeDeckFeed/FreeDeckFeed.tsx
```diff
@@ -29,7 +29,11 @@ function FreeDeckFeed() {
       return [...acc, ...page];
     }, []);
   }, [data]);
-  if (formattedData?.length === 0 && !isFetching) {
+
+  if (
+    (formattedData?.length === 0 || formattedData === undefined) &&
```
marvinmarnold, line 34: Is this related to this PR or an unrelated fix?

app/queries/home.ts
```diff
@@ -489,6 +489,7 @@ WHERE
     d."activeFromDate" IS NULL
     AND d."date" >= ${currentDayStart}
     AND d."date" <= ${currentDayEnd})
+    AND (c."hideDeckFromHomepage" = false OR c."hideDeckFromHomepage" IS NULL)
```
marvinmarnold, line 492: Normally I'd complain about lack of tests for this case. But we are in a grind right now.

marvinmarnold (approved):

app/components/FreeDeckFeed/FreeDeckFeed.tsx
yash431garg: It is related.

app/queries/home.ts
yash431garg: Okay, I will this in my backlog and will write the test whenever I will have some more bandwidth.

PR #902, opened 2025-02-01
Prod 712/box hub claim validation reward

yash431garg:
- Steps
1. Create a paid deck
2. Change the reveal Date
3. Add correct answer by revealing
4. Open mystery box

This PR doesn't account for.
1. Tests  
2. Mixpanel tracking


yash431garg:
cc: @samott 

It has the routes for rewards where we want to show history component.

app/queries/mysteryBox.ts
```diff
@@ -149,3 +163,363 @@ async function rewardTutorialMysteryBox(
...
+    });
+
+    const newMysteryBoxId = res.id;
+    return [...existingMysteryBoxIds, newMysteryBoxId];
+  }
+
+  return existingMysteryBoxIds;
+};
+
+export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
```
neel-ds, line 330: I think you should create new files for different functions

app/queries/mysteryBox.ts
```diff
@@ -149,3 +163,363 @@ async function rewardTutorialMysteryBox(
...
+    AND fatl."userId" = ${userId}
+    AND fatl."change" = -q."creditCostPerQuestion"
+    AND fatl."type" = 'PremiumQuestionCharge'
+    AND fatl."change" < 0;
+	`;
+
+  return filterQuestionsByMinimalNumberOfAnswers(questions);
+};
+
+export const rewardMysteryBoxHub = async ({}: {
```
neel-ds, line 244: new file here too

components/MysteryBox/MysteryBox.tsx
```diff
@@ -29,7 +29,7 @@ import MysteryBoxOverlay from "./MysteryBoxOverlay";
 
 const Lottie = dynamic(() => import("lottie-react"), { ssr: false });
 
-function buildMessage(lines: string[]) {
+export function buildMessage(lines: string[]) {
```
neel-ds, line 32: Can you create new util for it?

components/MysteryBox/MysteryBox.tsx
```diff
@@ -42,7 +42,7 @@ function buildMessage(lines: string[]) {
   );
 }
 
-type MysteryBoxStatus = "Idle" | "Opening" | "Closing";
+export type MysteryBoxStatus = "Idle" | "Opening" | "Closing";
```
neel-ds, line 45: It would be nice to move this to types/

neel-ds (changes requested):

samott:
Functionally, looks good.

- The animation is the Chompmas one.
- The "Mystery Boxes" title isn't centered:
mysterybox_title

app/components/MysteryBoxHub/MysteryBoxHub.tsx
```diff
@@ -0,0 +1,48 @@
+"use client";
```
samott, line 1: This should go in /components with the other MysteryBox components rather than /app/components,

samott:
I added in the history feed to the hub page on this branch, FYI.

app/queries/mysteryBox.ts
```diff
@@ -149,3 +163,363 @@ async function rewardTutorialMysteryBox(
...
+    AND EXISTS (
+        SELECT 1
+        FROM public."QuestionOption" qo
+        JOIN public."QuestionAnswer" qa ON qo.id = qa."questionOptionId"
+        WHERE 
+            qo."questionId" = q.id
+            AND qa.selected = TRUE
+            AND (qo."calculatedIsCorrect" IS NOT NULL OR qo."calculatedAveragePercentage" IS NOT NULL)
+            AND qa."userId" = ${userId}
+    )
```
samott, line 234: I think this will only work for subjective questions, right?

yash431garg:
> Functionally, looks good.
> 
> * The animation is the Chompmas one.
> * The "Mystery Boxes" title isn't centered:
> 
> mysterybox_title

I used Chompmas animation because it wasn't specific for credit and bonk. I will center the title

components/MysteryBox/MysteryBox.tsx
yash431garg: Sure

yash431garg:
> I added in the history feed to the hub page on this branch, FYI.

Thanks it slip from my mind

app/queries/mysteryBox.ts
yash431garg: You are correct. I will add a fix for this.

yash431garg: No, this query make sure that user only get a reward if the correct answer is calculated which means if you want to get more boxes you need to either do reveal or run the cron to calculate the answer.

components/MysteryBoxHub/OpenMysteryBox.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+function OpenMysteryBox({
+  isOpen,
+  closeBoxDialog,
+  boxType = EMysteryBoxCategory.Validation,
+  mysteryBoxIds,
+}: OpenMysteryBoxProps) {
+  const [status, setStatus] = useState<MysteryBoxStatus>("Idle");
+  const [isSubmitting, setIsSubmitting] = useState<boolean>(false);
+  const [MysteryBoxReward, setMysteryBoxReward] = useState<{
```
neel-ds, line 37: camel case

neel-ds (commented):
I think that openMysteryBoxHub fits better in actions instead of app/queries. There is typo in getValidationRewardQuestion filename. Noticed some inconsistencies in casing, like BuildMessage it would be good to keep camel case.

components/MysteryBoxHub/OpenMysteryBox.tsx
yash431garg: Thanks for pointing out these. I will fix them meanwhile BuildMessage behave as a component that's why I kept it that way.

yash431garg: Also, If you find more inconsistencies let me know happy to fix them.

app/queries/mysteryBox.ts
samott: yeah, but what about cases where we know the answer already (i.e. we use `isCorrect` instead of `calculatedIsCorrect`)? Or are we not giving boxes in that case?

samott: Okay, makes sense. Thanks for clarifying!

samott (approved):

neel-ds (approved):

PR #903, opened 2025-02-02
PROD-730: Cronjob to calculate correct answers for questions that are ready to reveal

samott:
Cronjob to calculate correct answers for questions that are ready to reveal.

### DIFF
--- a/__tests__/queries/answers.ts
+++ b/__tests__/queries/answers.ts
@@ -1,0 +1,203 @@
+import { getQuestionsNeedingCorrectAnswer } from "@/app/queries/answers";
+import prisma from "@/app/services/prisma";
+import { generateUsers } from "@/scripts/utils";
+import { QuestionType, Token } from "@prisma/client";
+import dayjs from "dayjs";
+
+describe("Answers", () => {
+  let deckId: number;
+  let questionIds: number[] = [];
+  let otherUsers: { id: string; username: string }[] = [];
+  let existingQuestionIds = {};
+  let questionNeedingAnswer: number;
+
+  beforeAll(async () => {
+    const futureDate = dayjs().add(1, "day").toDate();
+    const pastDate = dayjs().subtract(1, "day").toDate();
+
+    const existingQuestions = await prisma.question.findMany({
+      select: {
+        id: true,
+      },
+    });
+
+    existingQuestionIds = Object.fromEntries(
+      existingQuestions.map((question) => [question.id, true]),
+    );
+
+    await prisma.$transaction(async (tx) => {
+      // Create deck
+      const deck = await tx.deck.create({
+        data: {
+          deck: "Deck 1",
+          date: new Date(),
+          revealAtDate: futureDate,
+        },
+      });
+
+      deckId = deck.id;
+
+      // Create questions for decks
+      const questions = await Promise.all([
+        // Past question (ready to reveal), already have an answer
+        tx.question.create({
+          data: {
+            question: "Is the sky blue?",
+            type: QuestionType.BinaryQuestion,
+            revealAtDate: pastDate,
+            revealToken: Token.Bonk,
+            revealTokenAmount: 5000,
+            questionOptions: {
+              createMany: {
+                data: [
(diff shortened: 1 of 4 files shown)


app/queries/answers.ts
```diff
@@ -0,0 +1,51 @@
+import prisma from "@/app/services/prisma";
+
+export async function getQuestionsNeedingCorrectAnswer() {
```
yash431garg, line 3: We can give a better name something like getQuestionReadyToReveal. I think it already exist as well. https://github.com/gator-labs/chomp/blob/65a2a0e1c6d9d044776a0048990a24b38ea8b5d3/app/queries/home.ts#L239

app/queries/answers.ts
```diff
@@ -0,0 +1,51 @@
...
+      WHERE
+        ("calculatedIsCorrect" IS NULL OR "calculatedAveragePercentage" IS NULL)
+      GROUP BY q.id
+    )
+    WHERE (
+      (
+        ("revealAtDate" IS NOT NULL AND "revealAtDate" < NOW())
+        OR ("revealAtAnswerCount" IS NOT NULL AND "revealAtAnswerCount" >= "answerCount")
+      ) AND (
+        "answerCount" >= ${Number(process.env.MINIMAL_ANSWERS_PER_QUESTION ?? 3)}
```
yash431garg, line 47: I think if env doesn't exist we should through an error may be. What do you think?

app/api/cron/calculate-correct-answers/route.ts
```diff
@@ -0,0 +1,38 @@
+import { getQuestionsNeedingCorrectAnswer } from "@/app/queries/answers";
```
yash431garg, line 1: Does this means that a question will be delayed between 1-5min before ready to reveal?

yash431garg (changes requested):

app/queries/answers.ts
samott: I don't think we should throw a runtime error due to missing config.

I'm not a fan of having default values scattered throughout the code like this but this pattern is already established.

app/queries/answers.ts
samott: I'm open to using a different name, but ready to reveal != needing answer.

app/api/cron/calculate-correct-answers/route.ts
samott: Yes.

neel-ds (dismissed):

app/queries/answers.ts
yash431garg: Can you change this so that we can eventually remove this pattern?

app/queries/answers.ts
yash431garg: Can you adding comments for this?

app/queries/answers.ts
samott: I changed it for now. The correct, long-term solution is to check the config on startup.

__tests__/queries/answers.ts
```diff
@@ -0,0 +1,203 @@
...
+  let questionIds: number[] = [];
+  let otherUsers: { id: string; username: string }[] = [];
+  let existingQuestionIds = {};
+  let questionNeedingAnswer: number;
+
+  beforeAll(async () => {
+    const futureDate = dayjs().add(1, "day").toDate();
+    const pastDate = dayjs().subtract(1, "day").toDate();
+
+    const existingQuestions = await prisma.question.findMany({
```
neel-ds, line 18: Instead of ruling out all the questions, can't it just consider questions having revealAt > NOW()? 

It would be quick.

neel-ds (approved):

__tests__/queries/answers.ts
samott: > Instead of ruling out all the questions, can't it just consider questions having revealAt > NOW()?
> 
> It would be quick.

Sure, that's a valid approach. I'll leave it as is for now to get this over the line.

PR #904, opened 2025-02-02
PROD-709: Remove Chompmas box campaign

samott:
Removes code related to Chompmas mystery box campaign.


yash431garg (approved):
Is it fine to completely remove the code? What if we need this in future?

marvinmarnold (approved):

samott:
> Is it fine to completely remove the code? What if we need this in future? Do we need to remove allowlist table or it is fine to keep it?

It would be nice to keep it but there is a good argument for removing dead code, which increases bug surface area and code complexity. We have the code in the repo history at least in case we need to resurrect it.

The allowlist table is linked to `MysteryBoxPrize` so we should keep it.

PR #905, opened 2025-02-02
Fix: Allow for mystery boxes with no BONK in them (resolves failing tests)

samott:
The mechanism engine returns boxes without BONK sometimes, which makes the current code choke.

### DIFF
--- a/lib/mysteryBox.ts
+++ b/lib/mysteryBox.ts
@@ -124,8 +124,6 @@ export async function rewardMysteryBox(
       throw new Error(`Unimplemented trigger type: ${triggerType}`);
     }
 
-    if (!calculatedReward?.bonk) throw new Error("No BONK in mystery box");
-
     const userWallet = await prisma.wallet.findFirst({ where: { userId } });
 
     if (!userWallet) return null;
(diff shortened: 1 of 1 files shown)


yash431garg (approved):

lib/mysteryBox.ts
```diff
@@ -124,23 +124,23 @@ export async function rewardMysteryBox(
...
 
     if (!userWallet) return null;
 
     const tokenAddress = process.env.NEXT_PUBLIC_BONK_ADDRESS ?? "";
 
-    const prizes: MysteryBoxPrize[] = [
-      {
+    const prizes: MysteryBoxPrize[] = [];
+
+    if (calculatedReward?.bonk) {
```
marvinmarnold, line 135: I think `0` bonk will evaluate to false. In that case, no prize will be added. As opposed to a prize with 0 bonk as before?

I don't have a preference, just checking.

marvinmarnold (approved):

lib/mysteryBox.ts
samott: Yes, that was my intention. I have a slight preference for this approach because otherwise we have an asymmetry with credits. On the other hand, I could see an advantage to always having a prize row for each box.

PR #906, opened 2025-02-03
PROD-704: Remove mystery box after reveal all

samott:
Removes the reveal all / claim all mystery box.


yash431garg (approved):

neel-ds (changes requested):
I think there is more clean up require in useReveal hook and some other files. Following PR added the changes to move Mystery Box to Reveal All:

Ref: https://github.com/gator-labs/chomp/pull/823/files#diff-1d99ed0b5752b831f1d71008e07495d3cf637448c360127b43464f18defbc994

samott:
> I think there is more clean up require in useReveal hook and some other files. Following PR added the changes to move Mystery Box to Reveal All:

You're right - I've taken out the code around the reveal hook. There may be a few things left on the server side but I don't want to disrupt with the ongoing mystery box work.

neel-ds (dismissed):

neel-ds:
> > I think there is more clean up require in useReveal hook and some other files. Following PR added the changes to move Mystery Box to Reveal All:
> 
> You're right - I've taken out the code around the reveal hook. There may be a few things left on the server side but I don't want to disrupt with the ongoing mystery box work.

I think you can remove `getRevealAllMysteryBoxForQuestions` and related test as well.

PR #907, opened 2025-02-03
PROD-724: Add "credit cost per question" field to admin panel.

samott:
Add "credit cost per question" field to admin panel.

ccpq


yash431garg (approved):
LGTM haven't tested in preview.

neel-ds (approved):

PR #917, opened 2025-02-05
Prod 732/associate prize with trigger

yash431garg:
WIP -- I need to update all MB queries. As I have removed relation between MB and MBP

### DIFF
--- a/prisma/migrations/20250205092511_update_mystery_box_prize_trigger_relation/migration.sql
+++ b/prisma/migrations/20250205092511_update_mystery_box_prize_trigger_relation/migration.sql
@@ -1,0 +1,15 @@
+/*
+  Warnings:
+
+  - You are about to drop the column `mysteryBoxId` on the `MysteryBoxPrize` table. All the data in the column will be lost.
+
+*/
+-- DropForeignKey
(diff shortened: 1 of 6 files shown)


prisma/migrations/20250205092511_update_mystery_box_prize_trigger_relation/migration.sql
```diff
@@ -0,0 +1,48 @@
+-- Step 1: Add the new value to the enum type
+ALTER TYPE "EBoxTriggerType" ADD VALUE 'CombinedTriggers';
```
marvinmarnold, line 2: `CombinedTriggers` not named consistently with other existing triggers: [ClaimAllCompleted, DailyDeckCompleted, ChompmasStreakAttained, TutorialCompleted, RevealAllCompleted]

Just `Combined` is better.

marvinmarnold (dismissed):
I only looked at the migration code. It looks good to me. Let's get this merged as soon as tests are passing.

samott (commented):
FYI: I think the migrations should run after the ones I recently added:
20250206234031_add_chaintx_to_mystery_box_prize_{1,2,3}

So I think the date should be updated so the migration runs after mine -- or update my migration script (specifically 20250206234031_add_chaintx_to_mystery_box_prize_2)

PR #919, opened 2025-02-05
PROD-743: reduce wait period to reveal questions

samott:
Reduce correct answer cron interval to 1 minute and protect against multiple concurrent invocations.


app/api/cron/calculate-correct-answers/route.ts
```diff
@@ -34,5 +52,7 @@ export async function GET(request: Request) {
   } catch (error) {
     console.error("Error:", error);
     return new Response("Internal Server Error", { status: 500 });
```
yash431garg, line 54: I think we should release the mutex if this fails

yash431garg (changes requested):

app/api/cron/calculate-correct-answers/route.ts
samott: It will be released in the `finally` block.

yash431garg (approved):

PR #929, opened 2025-02-08
PROD-715: Optimize ATA Create Tx

neel-ds:
- Description
Add fixed CU with buffer proportion for receiver ATA creation

Test with new wallet

https://solscan.io/tx/jzLwGSa2VPbvnskGvKNheAKDgxs66qAbQZ9sDtw8dciyk8ySANfkPz726r8kegqnqmd9w62Q6AcbVtppUvFwVBV

### DIFF
--- a/app/utils/claim.ts
+++ b/app/utils/claim.ts
@@ -112,7 +112,7 @@ export const sendBonk = async (
   } else {
     // based on recent ATA creation CU consumption
     // https://solscan.io/tx/37aEx5VpNMLXrKrMYPoGFbUmhg2YBGgvR3azHY17mWk1uupve229jhdyJWxA3HHAuJq8mtmeQBSR7dkuvhnVYbgs
-    const computeUnitFix = 27000;
+    const computeUnitFix = 27695;
 
     // Buffer to compensate any additional usage
     const computeUnitsIx = ComputeBudgetProgram.setComputeUnitLimit({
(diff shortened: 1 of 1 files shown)


neel-ds:
Tested on preview: https://solscan.io/tx/jzLwGSa2VPbvnskGvKNheAKDgxs66qAbQZ9sDtw8dciyk8ySANfkPz726r8kegqnqmd9w62Q6AcbVtppUvFwVBV

app/utils/claim.ts
```diff
@@ -108,6 +108,17 @@ export const sendBonk = async (
     });
 
     // update the instructions with compute unit instruction (unshift will move compute unit to the start and it is recommended in docs as well.)
+    instructions.unshift(computeUnitsIx);
+  } else {
+    // based on recent ATA creation CU consumption
+    // https://solscan.io/tx/37aEx5VpNMLXrKrMYPoGFbUmhg2YBGgvR3azHY17mWk1uupve229jhdyJWxA3HHAuJq8mtmeQBSR7dkuvhnVYbgs
```
marvinmarnold, line 114: The sample tx used 27,695, this only uses 27,000. Line 103 uses a 10% buffer, this is using 5%. Any reason you decided to go with this, smaller buffer?

neel-ds: Above mentioned tested tx consumed 25325 CU approx. and code sample L114 was tested by Yash in Phantom (not by our code). There is gap between both consumption number so I went for smaller buffer to cater both range.

PR #930, opened 2025-02-08
PROD-689: Automated banned users

samott:
Pulls suspected bot users from the mechanism engine and updates the database.

Runs 10 minutes before every hour so questions that are about to reveal are included in the analysis.

The production endpoint is not deployed yet.

### DIFF
--- a/app/api/cron/detect-bot-activity/route.ts
+++ b/app/api/cron/detect-bot-activity/route.ts
@@ -47,7 +47,11 @@ export async function GET(request: Request) {
     const scores: Record<string, number> = results["mean_bot_score"];
     const botUserIds = Object.keys(scores);
 
-    await updateBots(botUserIds);
+    await updateBots(
+      botUserIds,
+      new Date(results.start_date),
+      new Date(results.end_date),
+    );
 
     return new Response(
       JSON.stringify({
--- a/app/api/cron/detect-bot-activity/route.ts
+++ b/app/api/cron/detect-bot-activity/route.ts
@@ -1,5 +1,7 @@
+import { SENTRY_FLUSH_WAIT } from "@/app/constants/sentry";
 import { tryAcquireMutex } from "@/app/utils/mutex";
 import { updateBots } from "@/lib/bots";
+import * as Sentry from "@sentry/nextjs";
 
 const API_TIMEOUT = 5 * 60 * 1000; // Five minutes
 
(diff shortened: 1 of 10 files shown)


lib/jwt.ts
```diff
@@ -5,7 +5,7 @@ import { cookies } from "next/headers";
 export const checkThreatLevel = async (userId: string) => {
   const user = await prisma.user.findFirst({ where: { id: userId } });
   if (!user) throw new Error("User not found");
-  if (user.threatLevel) {
+  if (user.threatLevel === "bot") {
```
marvinmarnold, line 8: We could make this a constant/enum.

lib/bots.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param bots Array of user IDs.
+ */
+export const updateBots = async (bots: string[]) => {
+  await prisma.user.updateMany({
+    data: {
+      threatLevel: "bot",
+    },
+    where: {
+      id: { in: bots },
+      threatLevel: null,
```
marvinmarnold, line 16: I know I said we should prevent overwriting users who have been pardoned. I don't think we should. 

We just need to make sure that in the future if we pardon someone that we don't pardon them until after the window that mechanism engine looks at has passed. Otherwise the same user will be immediately re-identified as a bot.

Scope increase, nice to have:
Ideally, the case where a pardoned user is detected botting again could throw a Sentry event that we could then turn into a Slack alert with clear message about the username, address, and userId suspected of being a bot.

marvinmarnold (approved):

yash431garg (dismissed):

marvinmarnold:
I just merged the mechanism code for this fyi

samott:
@marvinmarnold I updated the code with more features. The code automatically checks if it's changed the threat level in the current analysis window (I updated the mechanism engine to output the dates used), and there are a few different enums at our disposal to control re-blocking behaviour.

app/api/cron/detect-bot-activity/route.ts
```diff
@@ -0,0 +1,72 @@
...
+        message: "Analysis completed",
+        botCount: botUserIds.length,
+      }),
+      {
+        status: 200,
+        headers: { "Content-Type": "application/json" },
+      },
+    );
+  } catch (error) {
+    console.error("Error:", error);
```
marvinmarnold, line 67: I believe error logs get picked up by Sentry but I can't remember. The most likely reason this will happen is if mechanism-engine runs out of memory. Can we do Sentry.captureMessage here instead and make it more explicit what the error is?

marvinmarnold (commented):
Code looks good. But some of the test failures look like they are related to this code.

PR #944, opened 2025-02-11
fix: only show mystery box with a prize in history tab

yash431garg:
Problem: 
There were some card in history page with 0bonk, 0credit and with unopened state.

Issue: 
For triggers like ClaimAllCompleted/RevealAllCompleted we don't have any trigger to prize to relation now. Instead it is now with triggerType Combined to prize. But fetchHistory query returned the trigger without prize cause the corrupt data. 

Solution:
Updated query for fetch history to show data if prize is available for a trigger. 

Extra: 
1. Fixed test
2. Show total mystery box reward
3. Show tag for validation reward.

### DIFF
--- a/app/actions/mysteryBox/fetchHistory.ts
+++ b/app/actions/mysteryBox/fetchHistory.ts
@@ -26,6 +26,9 @@ export async function fetchMysteryBoxHistory({
       triggers: {
         some: {
           triggerType: {},
+          MysteryBoxPrize: {
+            some: {}, // Ensures there is at least one MysteryBoxPrize
+          },
         },
       },
     },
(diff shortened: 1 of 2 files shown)


app/actions/mysteryBox/fetchHistory.ts
```diff
@@ -65,21 +68,27 @@ export async function fetchMysteryBoxHistory({
...
-      else if (prize.prizeType == "Token" && prize.tokenAddress == bonkAddress)
-        bonkReceived = prize.amount;
+
+      if (prize.prizeType == "Credits") {
+        creditsReceived += parseFloat(prize.amount); // Sum the credits amount
+      } else if (
+        prize.prizeType == "Token" &&
+        prize.tokenAddress == bonkAddress
+      ) {
+        bonkReceived += parseFloat(prize.amount); // Sum the bonk amount
```
samott, line 78: Ideally we should use Decimal.js to do the addition.

samott: Actually, for BONK we are not doing it elsewhere so this is ok for now.

samott (approved):

PR #950, opened 2025-02-11
PROD-765: mystery box show as claimable but nothing added to history

samott:
This fixes the mystery box loop I was seeing. There were a couple of issues: the action was returning unopened boxes, and returning the same box multiple times (for each trigger) - so I think having multiple questions in a box was the problem.

### DIFF
--- a/app/actions/mysteryBox/openMysteryBoxHub.ts
+++ b/app/actions/mysteryBox/openMysteryBoxHub.ts
@@ -54,7 +54,7 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
           in: mysteryBoxIds,
         },
         userId,
-        status: EMysteryBoxStatus.New,
+        status: { in: [EMysteryBoxStatus.New, EMysteryBoxStatus.Unopened] },
       },
 
       include: {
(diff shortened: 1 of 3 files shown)


app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -206,39 +206,6 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
 
     release();
   } catch (e) {
-    await prisma.mysteryBox.updateMany({
```
yash431garg, line 209: Could you explain why you removed this?

samott: Because I think we'll only be in the exception handler if the database transaction fails (in which case there's nothing to update) or the BONK transaction failed, in which case we won't have updated the database anyway.

yash431garg: But we have three states New, Unclaimed, claimed. I thought we will use unclaimed to denote that there is some error in reward distribution.

samott: Ok - fair enough. I've put back this logic. However, I also updated the code to allow unopened boxes to be included in the next attempt (otherwise they get stuck).

marvinmarnold (approved):

PR #955, opened 2025-02-12
PROD-788: Update SOL Payment Address

neel-ds:
Replaces the treasury address with different env of address to accept sol payment.

NOTE: `NEXT_PUBLIC_SOL_PAYMENT_ADDRESS` env is required for this deployment.

Includes patch for failing tests


samott (commented):
I feel like there is likely to be some confusion with the names here since the original treasury still exists. Maybe this should be called `SOL_PAYMENT_DESTINATION` or something?

PR #961, opened 2025-02-13
fix: credit purchase

yash431garg:


actions/credits/createChainTx.ts
```diff
@@ -1,12 +1,11 @@
-"use server";
-
 import { getTreasuryAddress } from "@/actions/getTreasuryAddress";
```
neel-ds, line 1: I think it would be better to move this file from actions/ to lib/

actions/credits/updateTxStatusConfirm.ts
```diff
@@ -1,5 +1,3 @@
-"use server";
-
 import { getJwtPayload } from "@/app/actions/jwt";
```
neel-ds, line 1: same move to lib/

actions/credits/verifyPayment.ts
```diff
@@ -1,5 +1,3 @@
-"use server";
-
 import { getJwtPayload } from "@/app/actions/jwt";
```
neel-ds, line 1: move to lib/

lib/credits/processTransaction.ts
```diff
@@ -37,12 +35,9 @@ export async function processTransaction(
   }
 
   // Send transaction
-  const txHash = await CONNECTION.sendRawTransaction(
-    signedTransaction.serialize(),
-    {
-      skipPreflight: true,
-    },
-  );
+  const txHash = await CONNECTION.sendEncodedTransaction(signedTransaction, {
```
neel-ds, line 38: I think this should be wrapped in a try-catch or validated upfront. If a user invokes initiatePurchase with a valid signature and an already on-chain transaction, it will fail and needs proper error handling.

lib/credits/initiatePurchase.ts
```diff
@@ -1,24 +1,25 @@
+"use server";
+
 import { createSignedSignatureChainTx } from "@/actions/credits/createChainTx";
```
neel-ds, line 3: move to /actions

neel-ds (changes requested):
The file structure needs some refactoring, and tests should reflect those changes. The test is failing—potentially, breaking down the Prisma transaction in deleteDeck could resolve it.

lib/credits/processTransaction.ts
yash431garg: It will fail in confirmation

neel-ds: I don't think so that confirmation will fail if valid on-chain transaction is provided

yash431garg: If a use invoke this with a valid signedTransaction It will return the signature for that. But It will fail blockhash is expired.

neel-ds: sendEncodedTransaction will fail because transaction is already on-chain and confirmation would work either way because we provide latest blockhash to verify the commitment.

yash431garg: Convinced

neel-ds (approved):
LGTM!

samott (approved):
Looks good.

PR #965, opened 2025-02-13
PROD-770: Credits bulk buy modal

samott:
Implements credits bulk buy modal.

deckscreen
modal
profile


app/components/Icons/CoinStackIcon.tsx
```diff
@@ -0,0 +1,36 @@
+import { IconProps } from ".";
```
neel-ds, line 1: This icon already exists with name CoinsIcon file

neel-ds (commented):
I think yarn lockfile update is missing as there is change in package json.

yash431garg (dismissed):
LGTM other than failing workflow

PR #967, opened 2025-02-14
fix: enforce sufficient credits to view question | PROD-791

marvinmarnold:
- Description: Prevent questions from being viewed without paying credits. Only allow viewed questions to be answered.
- What are the steps to test that this code is working? Invoke question answering via terminal. Normal question answering via UI should still work.

- Moved relevant actions to their own files.
- Moved some actions to normal lib files, if they don't need to be accessed via frontend.
- Throw error if viewing question would lead to negative balance.
-


neel-ds (commented):
answerQuestion action should validate credit payment from fatl before storing answers

neel-ds (approved):

PR #968, opened 2025-02-14
feat: remove bot response from answer calculation

yash431garg:
Description
Intended behavior is to not include bot response in correct answer  calculation and reward calculation
Solution
Updated answerPercentage query to calculate correct answer data based on non bots user responses.

### DIFF
--- a/__tests__/utils/algo/calculate-correct-answer.test.ts
+++ b/__tests__/utils/algo/calculate-correct-answer.test.ts
@@ -238,6 +238,9 @@ describe("calculateCorrectAnswe", () => {
     const questionOptionsBeforeMarkedBot = await prisma.questionOption.findMany(
       {
         where: {
+          calculatedAveragePercentage: {
+            not: null,
+          },
           question: {
             deckQuestions: {
               some: {
@@ -264,6 +267,9 @@ describe("calculateCorrectAnswe", () => {
     await calculateCorrectAnswer([deckIds[0]]);
     const questionOptionsAfterMarkedBot = await prisma.questionOption.findMany({
       where: {
+        calculatedAveragePercentage: {
+          not: null,
+        },
         question: {
           deckQuestions: {
             some: {
@@ -292,6 +298,9 @@ describe("calculateCorrectAnswe", () => {
 
     const questionOptionsAfterMarkedBot = await prisma.questionOption.findMany({
       where: {
+        calculatedAveragePercentage: {
+          not: null,
+        },
         question: {
           deckQuestions: {
             some: {
@@ -320,6 +329,9 @@ describe("calculateCorrectAnswe", () => {
     const questionOptionsBeforeMarkedBot = await prisma.questionOption.findMany(
       {
         where: {
+          calculatedAveragePercentage: {
+            not: null,
+          },
           question: {
             deckQuestions: {
               some: {
--- a/__tests__/utils/algo/calculate-correct-answer.test.ts
+++ b/__tests__/utils/algo/calculate-correct-answer.test.ts
@@ -1,0 +1,346 @@
+// TODO: 1. Write a test to check if what will be the result if we exclude the bots in binary question
+// 2. Write a test to check if what will be the result if we exclude the bots in multiple choice question
+import { deleteDeck } from "@/app/actions/deck/deck";
+import prisma from "@/app/services/prisma";
+import { calculateCorrectAnswer } from "@/app/utils/algo";
+import { generateUsers } from "@/scripts/utils";
+import { EThreatLevelType } from "@/types/bots";
+import { QuestionType } from "@prisma/client";
+
+jest.mock("next/headers", () => ({
+  headers: jest.fn(() => ({
+    get: jest.fn((key) => {
+      if (key === "x-path") return "/some-path";
+      return null;
+    }),
+  })),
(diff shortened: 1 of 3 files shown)


samott (commented):
In the ticket, it's stated that already-revealed answers shouldn't affected. Not sure if there is a reasonable approach to that requirement though since marking a new bot will affect all their previous answers...

yash431garg:
> In the ticket, it's stated that already-revealed answers shouldn't affected. Not sure if there is a reasonable approach to that requirement though since marking a new bot will affect all their previous answers...

Previously I thought that we are only using the updated query to calculate answer for a question which are fetched by this https://github.com/gator-labs/chomp/blob/fda4019b3e933940e6913e04c95b2790084fa9bc/app/api/cron/calculate-correct-answers/route.ts#L36

But I guess I had to double check as the answerPercentageQuery is used in other places as well.

__tests__/utils/algo/calculate-correct-answer.test.ts
```diff
@@ -0,0 +1,346 @@
...
+}));
+
+jest.mock("next/navigation", () => ({
+  redirect: jest.fn(),
+}));
+jest.mock("next/cache", () => ({
+  revalidatePath: jest.fn(),
+}));
+
+describe("calculateCorrectAnswe", () => {
```
neel-ds, line 30: Typo at last

__tests__/utils/algo/calculate-correct-answer.test.ts
```diff
@@ -0,0 +1,346 @@
...
+      data: users,
+    });
+
+    userIds = users.map((user) => user.id);
+
+    // create decks and questions
+    const decksData = [
+      {
+        data: {
+          deck: `Premium Deck ${new Date().getTime()}`,
```
neel-ds, line 46: Having distinct deck names would make debugging easier

__tests__/utils/algo/calculate-correct-answer.test.ts
```diff
@@ -0,0 +1,346 @@
...
+    const deletePromises = deckIds.map((deckId) => deleteDeck(deckId));
+    await Promise.all(deletePromises);
+
+    await prisma.user.deleteMany({
+      where: {
+        id: { in: userIds },
+      },
+    });
+  });
+  it("should compare binary question reslut for bots and non-bots user response", async () => {
```
neel-ds, line 235: typo result*

__tests__/utils/algo/calculate-correct-answer.test.ts
yash431garg: I will keep in mind from next time.

app/queries/answerPercentageQuery.ts
```diff
@@ -59,3 +63,32 @@ export async function answerPercentageQuery(questionOptionIds: number[]) {
     ),
   }));
 }
+
+export async function calculatedQuestionOrderPercentages(
```
marvinmarnold, line 67: Not obvious to me what this function is doing based on the name. Comments would be helpful and maybe a clearer name.

app/queries/question.ts
```diff
@@ -214,9 +214,21 @@ export async function getQuestionWithUserAnswer(questionId: number) {
...
+        qo.calculatedPercentageOfSelectedAnswers,
+      secondOrderAveragePercentagePicked: qo.calculatedAveragePercentage,
+    }),
+  );
+
   const questionOptionIds = question.questionOptions.map((qo) => qo.id);
-  const questionOptionPercentages =
-    await answerPercentageQuery(questionOptionIds);
+
+  const questionOptionPercentages = await calculatedQuestionOrderPercentages(
```
marvinmarnold, line 228: Should this be `const questionOrderPercentages`?

app/queries/answerPercentageQuery.ts
```diff
@@ -23,21 +24,24 @@ export async function answerPercentageQuery(questionOptionIds: number[]) {
                   (
                     select count(*)
                     from public."QuestionAnswer" subQa
-                    where subQa.selected = true and subQa."questionOptionId" = qo."id" and subQa."status" = 'Submitted'
+                    join public."User" u on subQa."userId" = u."id"
```
marvinmarnold, line 27: I don't want to prolong this task. In the future, you could consider making a DB view for `select * from Users u where (u."threatLevel" IS NULL OR u."threatLevel" IN (${EThreatLevelType.ManualAllow}, ${EThreatLevelType.PermanentAllow})`. 

If the view is named something like `AuthorizedUsers`, then this query could be rewritten as:

```sql
select count(*)
                    from public."QuestionAnswer" subQa
                    INNER join public."AuthorizedUser" u on subQa."userId" = u."id"
where subQa.selected = true and subQa."questionOptionId" = qo."id" and subQa."status" = 'Submitted'
```

Notice the extra "INNER". This approach would reduce code duplication and make it easier to read.

marvinmarnold (approved):

app/queries/answerPercentageQuery.ts
yash431garg: Thank you! I will keep this in mind for future work considering the current priority tasks.

marvinmarnold (approved):

neel-ds (approved):
LGTM!

PR #970, opened 2025-02-14
PROD-795: Not Allow Null Credit Cost 

neel-ds:
Admin will see 0 credit as default. Adding not null validation in schema would affect the legacy questions.

### DIFF
--- a/app/components/AdminBinaryQuestionGeneratorForm/AdminBinaryQuestionGeneratorForm.tsx
+++ b/app/components/AdminBinaryQuestionGeneratorForm/AdminBinaryQuestionGeneratorForm.tsx
@@ -80,7 +80,6 @@ export default function AdminBinaryQuestionGeneratorForm({
             required
             className="mt-1 block w-full px-4 py-2 border border-gray-600 rounded-md bg-gray-800 text-white placeholder-gray-400 focus:ring-indigo-500 focus:border-indigo-500"
           >
-            <option value="none">None (legacy)</option>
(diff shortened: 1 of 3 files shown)


samott (commented):
There's also `AdminMultiChoiceQuestionGeneratorForm.tsx` and `AdminBinaryQuestionGeneratorForm.tsx` FYI

samott (approved):

neel-ds:
> There's also `AdminMultiChoiceQuestionGeneratorForm.tsx` and `AdminBinaryQuestionGeneratorForm.tsx` FYI.
> 
> Although there is a case for leaving them: maybe useful for testing legacy handling in the future?

Ah I missed it. No problem, I will update it.

samott (approved):

marvinmarnold (commented):
Can't we add non-null validation here without impacting the DB? 

https://github.com/gator-labs/chomp/blob/c5be8ea8f2a89157e50b67f222e80193f815719c/app/schemas/deck.ts#L10

neel-ds:
> Can't we add non-null validation here without impacting the DB?
> 
> https://github.com/gator-labs/chomp/blob/c5be8ea8f2a89157e50b67f222e80193f815719c/app/schemas/deck.ts#L10

I tried to change it for deck as well as question but in DeckForm we use `getDeckSchema` which fetch data from database where CPQ can be null. It violates then.

marvinmarnold (approved):

PR #981, opened 2025-02-17
PROD-805: Auto populate revealTokenAmount input in admin panel

samott:
- Auto populate `revealTokenAmount` input in admin panel unless manually set.
- Send cost and reveal amounts to mechanism engine.

This update will need to be coordinated with the mechanism engine as we are changing a parameter's meaning and expected value by orders of magnitude.

### DIFF
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -243,6 +243,7 @@ export const calculateReward = async (
       second_order_mean: 0,
       second_order_estimates: [0],
       question_cost: 0,
+      token_reward: 0,
     };
 
     const correctOptionIndex = question.questionOptions.findIndex(
@@ -284,7 +285,8 @@ export const calculateReward = async (
           questionOption?.calculatedAveragePercentage ??
           getAverage(second_order_estimates),
         second_order_estimates,
-        question_cost: question.revealTokenAmount,
+        question_cost: question.creditCostPerQuestion ?? 0,
+        token_reward: question.revealTokenAmount,
       };
     }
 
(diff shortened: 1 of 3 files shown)


neel-ds (dismissed):
LGTM! Can we change the default revealTokenAmount wrt. credit (as default 0)?
Screenshot 2025-02-17 at 5 49 51 PM

neel-ds (dismissed):

app/utils/algo.ts
```diff
@@ -402,6 +405,11 @@ export const calculateMysteryBoxHubReward = async (
       )
       .find((answer) => answer.userId === userId && answer.selected);
 
+    if (question.creditCostPerQuestion === null)
```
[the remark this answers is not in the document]
samott: Yeah - I added the fallback because TS wasn't narrowing the type, but maybe moving it up will help.

marvinmarnold:
FYI @samott this is high priority. As soon as this is ready, I will cut a release and merge https://github.com/gator-labs/mechanism-engine/pull/27/files

app/utils/algo.ts
```diff
@@ -494,7 +509,8 @@ export const calculateMysteryBoxHubReward = async (
           questionOption?.calculatedAveragePercentage ??
           getAverage(second_order_estimates),
         second_order_estimates: second_order_estimates,
-        question_cost: question.revealTokenAmount,
+        question_cost: question.creditCostPerQuestion,
+        token_reward: question.revealTokenAmount,
       };
     }
```
marvinmarnold, line 516: Should revert to using `/rewards`, not `v2/rewards`

app/utils/algo.ts
```diff
@@ -390,6 +398,11 @@ export const calculateMysteryBoxHubReward = async (
   }[] = [];
 
   for (const question of questions) {
+    if (question.creditCostPerQuestion === null)
+      throw new Error(
```
marvinmarnold, line 402: I will scope out a new ticket so that errors with one question don't stop the process from working overall. Both for creditCostErrors but also any errors that might happen while invoking getMechanismEngineResponse

marvinmarnold: https://linear.app/gator/issue/PROD-822/mechanism-engine-failures-should-not-interfere-with-other-questions

PR #987, opened 2025-02-18
fix: update get new history query

yash431garg:
Issue: Previously we used to show incorrect tag even though user had given a correct answer. 
Screenshot 2025-02-18 at 5 40 59 PM

Resolution: This PR update the query parameters to show tags correctly.

### DIFF
--- a/__tests__/queries/history/get-new-history-query.test.ts
+++ b/__tests__/queries/history/get-new-history-query.test.ts
@@ -1,0 +1,214 @@
+import { deleteDeck } from "@/app/actions/deck/deck";
+import { getNewHistoryQuery } from "@/app/queries/history";
+import prisma from "@/app/services/prisma";
+import { generateUsers } from "@/scripts/utils";
+import { QuestionType } from "@prisma/client";
+
+jest.mock("next/headers", () => ({
+  headers: jest.fn(() => ({
+    get: jest.fn((key) => {
+      if (key === "x-path") return "/some-path";
+      return null;
+    }),
+  })),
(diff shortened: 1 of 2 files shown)


samott (commented):
There is a similar query in this file for the totals that needs to be updated too.

yash431garg:
> There is a similar query in this file for the totals that needs to be updated too.

I didn't get you. Can you refer me to the query?

samott:
> > There is a similar query in this file for the totals that needs to be updated too.
> 
> I didn't get you. Can you refer me to the query?

It's here. It uses the same logic:

https://github.com/gator-labs/chomp/blob/ef068e4258761424e46d2305b88f83e7098baa2b/app/queries/history.ts#L261

marvinmarnold (changes requested):
Add tests for any queries that change.

yash431garg:
> Add tests for any queries that change.

Yes, tests are in progress.

yash431garg:
> > > There is a similar query in this file for the totals that needs to be updated too.
> > 
> > 
> > I didn't get you. Can you refer me to the query?
> 
> It's here. It uses the same logic:
> 
> https://github.com/gator-labs/chomp/blob/ef068e4258761424e46d2305b88f83e7098baa2b/app/queries/history.ts#L260-L289

Thank you. I am a bit concerned that in the future these may slip from our minds. What would you suggest for handling similar queries?

yash431garg:
Failing test is unrelated and will be removed in @neel-ds PR

samott (approved):

PR #989, opened 2025-02-19
Prod 838/show campaign mystery box

yash431garg:
Description --> 
Add scaffold for disabled box and campaign box table


lib/getCampaignBoxes.ts
```diff
@@ -0,0 +1,13 @@
...
+import prisma from "@/app/services/prisma";
+import "server-only";
+
+export const getCampaignBoxes = async () => {
+  const payload = await getJwtPayload();
+
+  if (!payload) {
+    return null;
+  }
+
```
samott, line 11: Is the auth check really needed if this is a library function?

marvinmarnold (approved):

PR #995, opened 2025-02-20
PROD-798: Update connect wallet screen content

samott:
Update the connect wallet screen content.

connect_wallet


yash431garg:
Screenshot 2025-02-20 at 5 20 47 PM
Spacing looks bit odd to me.

marvinmarnold (commented):
Looks good. You used the right links but they aren's accessible. I asked the team about this on Linear.

yash431garg (dismissed):
Screenshot 2025-02-21 at 11 57 07 AM

I am not sure if it is based figma design. But I still feel that we can add spacing between button and above text like this

neel-ds (approved):
LGTM! I think justify-between would give better alignment.

samott:
The content was bunched up at the top in the copy - I do agree it looks better with dynamic spacing so I've updated the layout.

Not too sure if it's a good idea separating the legal text and the button, but it looks better (can maybe revisit when the feedback bar is implemented).

neel-ds:
> Visit Preview

I think button and legal text together looks fine. Dynamic gap between image <> content <> button + legal would look good.

marvinmarnold (approved):
I generally agree with all the comments about bunched up design. But that's on us @gator-labs/devs to review the issues more closely at the beginning of the cycle. It's generally bad practice for us to be making last minute changes to the designs.

marvinmarnold (changes requested):
[deleted comment]

marvinmarnold (approved):
Whoops, my last comment belonged in: https://github.com/gator-labs/chomp/pull/999

PR #997, opened 2025-02-21
Skip flaky tests

neel-ds:
home/deck.test.ts and get-stack.test.ts are flaky tests. We reviewed the logic thoroughly and fixed time zone issue but worked out. We can rewrite these tests later. Skipping it for now.


marvinmarnold (approved):
Can you make sure there is a follow up issues for this. These seem like important tests for us to have coverage for.

neel-ds:
> Can you make sure there is a follow up issues for this. These seem like important tests for us to have coverage for.

Sure, I created PROD-848 to follow up on this issue.

samott (approved):

PR #998, opened 2025-02-21
PROD-836: Fix Credit Transaction Verfication

neel-ds:
- Changed the sequence of instructions in Transaction
- Add verification of metadata sanity check


lib/credits/processTransaction.ts
```diff
@@ -77,9 +77,17 @@ export async function processTransaction(
         throw new Error("Payment could not be verified");
       }
 
+      const result = await CONNECTION.getParsedTransaction(txHash, {
```
marvinmarnold, line 80: I read your comment [here](https://linear.app/gator/issue/PROD-836/prevent-failed-transactions-from-being-confirmed#comment-70ba8c62), but I am still not clear on why we need `getParsedTransaction`.

1. Does `getParsedTransaction` wait until transaction is finalized?
2. How much latency does this add?

neel-ds: 1. Till it's confirmed
2. 3ms (nothing) This sanity check validates if transaction simulation fails on-chain then it will throw error.

lib/credits/processTransaction.ts
```diff
@@ -77,9 +77,17 @@ export async function processTransaction(
...
+        commitment: TRANSACTION_COMMITMENT,
+      });
+
+      if (!result || result?.meta?.err) {
+        throw new Error("Transaction failed");
```
marvinmarnold, line 85: We should wrap the original error so that we can get clearer info in logs and sentry.

marvinmarnold (changes requested):
Awesome, this approach looks good. Just a small comment.

marvinmarnold (approved):

lib/credits/processTransaction.ts
marvinmarnold: @neel-ds can you confirm that this won't result in two error events being sent to Sentry? One for Sentry.captureException. And another for when the thrown error is handled by the invoking function?

neel-ds: Yes, it was capturing two errors in the flow. I updated it to avoid repetition.

samott (approved):

yash431garg (approved):
LGTM... Is this solution here to check result.meta.err for confirmation?

neel-ds:
> LGTM... Is this solution here to check result.meta.err for confirmation?

yes

PR #1005, opened 2025-02-24
PROD-849: Fix Open MysteryBox Hub

neel-ds:
- openMysteryBoxHub query validate that MBP should be unclaimed or dismissed
- It will allow new or unopened MB with unclaimed or dismissed MBP
- Better Error Handling in openMysteryBoxHub action and sendBonk util
- Fix FATL update query to handle values
- Add retry logic if prisma tx API fails
- Add similar sanity check on transaction confirmation verification like credit payment to capture tx simulation failure

Few changes won't be related but I changed utils/claim.ts --> utils/sendBonk.ts so legacy mysteryBox code is refactored.

### DIFF
--- a/app/actions/mysteryBox/openMysteryBoxHub.ts
+++ b/app/actions/mysteryBox/openMysteryBoxHub.ts
@@ -128,7 +128,6 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
     if (totalBonkAmount > 0) {
       txHash = await sendBonkFromTreasury(totalBonkAmount, userWallet.address);
       if (!txHash) {
-        release();
         throw new Error("Send bonk transaction failed");
       }
     }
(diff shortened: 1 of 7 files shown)


samott (approved):

app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -124,132 +128,148 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
     if (totalBonkAmount > 0) {
       txHash = await sendBonkFromTreasury(totalBonkAmount, userWallet.address);
       if (!txHash) {
-        release();
-        throw new Error("Tx failed");
+        throw new Error("Send bonk transaction failed");
```
marvinmarnold, line 131: Can we add more info to the error?

app/utils/sendBonk.ts
```diff
@@ -73,7 +72,7 @@ export const sendBonk = async (
 
   instructions.push(transferInstruction);
 
-  let blockhashResponse = await CONNECTION.getLatestBlockhash("confirmed");
+  let blockhashResponse = await CONNECTION.getLatestBlockhash("finalized");
```
marvinmarnold, line 75: We've been trying to avoid waiting for finalization. Why did you side to do this now.

marvinmarnold (approved):

PR #1006, opened 2025-02-25
feat: script to fix NaN prizes | PROD-854

marvinmarnold:
- Description: Find all MysteryBoxPrize where amount=NaN and recalculate the rewards
- What are the steps to test that this code is working? `yarn dev:fix-nan-mystery-box`

### DIFF
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -1,5 +1,4 @@
 import { QuestionType } from "@prisma/client";
-import "server-only";
 
 import { answerPercentageQuery } from "../queries/answerPercentageQuery";
 import prisma from "../services/prisma";
--- a/scripts/fix-nan-mystery-box.ts
+++ b/scripts/fix-nan-mystery-box.ts
@@ -1,0 +1,102 @@
+import { EBoxPrizeType, PrismaClient } from "@prisma/client";
+import Decimal from "decimal.js";
+
+import { calculateMysteryBoxHubReward } from "../app/utils/algo";
+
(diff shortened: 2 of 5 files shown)


app/utils/algo.ts
```diff
@@ -1,5 +1,4 @@
 import { QuestionType } from "@prisma/client";
-import "server-only";
```
marvinmarnold, line 2: @yash431garg I'm not sure why I had to remove this. Otherwise I wasn't able to run the script because it would say I was trying to run from a client component.

yash431garg: May be you should create a separate utility for this purpose because it contains some important utility and we may end up importing them in client which can cause us trouble.

marvinmarnold: I think the issue was just importing 'server-only' on the first line. It's working now.

scripts/fix-nan-mystery-box.ts
```diff
@@ -0,0 +1,102 @@
...
+import { calculateMysteryBoxHubReward } from "../app/utils/algo";
+
+const prisma = new PrismaClient();
+
+async function updateMysteryBoxPrizes() {
+  // Step 1: Query for all MysteryBoxPrizes where amount = NaN
+  const mysteryBoxPrizes = await prisma.mysteryBoxPrize.findMany({
+    where: {
+      amount: "NaN", // We're looking for "NaN" stored as a string
+      prizeType: EBoxPrizeType.Token,
```
marvinmarnold, line 13: Remove

PR #1016, opened 2025-02-28
PROD-855: Update streaks to consider credit purchases and mystery box claims.

samott:
Update streaks to consider credit purchases and mystery box claims.

### DIFF
--- a/app/queries/home.ts
+++ b/app/queries/home.ts
@@ -298,10 +298,11 @@ async function queryUsersLatestStreak(userId: string): Promise<number> {
     AND qa."status" = 'Submitted'
     UNION
     SELECT DISTINCT DATE("createdAt") AS activityDate
-    FROM public."fungibleAssetTransactionLog" fatl
+    FROM public."FungibleAssetTransactionLog" fatl
     WHERE "userId" = ${userId}
     AND fatl."asset" = 'Credit'
     AND fatl."type" = 'CreditPurchase'
+    UNION
     SELECT DISTINCT DATE("createdAt") AS activityDate
     FROM public."MysteryBox" mbox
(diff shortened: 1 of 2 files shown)


app/queries/home.ts
```diff
@@ -290,27 +290,37 @@ async function queryUsersLatestStreak(userId: string): Promise<number> {
...
     SELECT DISTINCT DATE("createdAt") AS activityDate
     FROM public."ChompResult"
-    WHERE "userId" = ${userId}  
+    WHERE "userId" = ${userId}
     UNION
     SELECT DISTINCT DATE("createdAt") AS activityDate
     FROM public."QuestionAnswer" qa
     WHERE "userId" = ${userId}
     AND qa."status" = 'Submitted'
+    UNION
```
marvinmarnold, line 299: General comment, not uniquely related to this PR. Something to think about. I'm not suggesting any immediate action.

It's hard for me to review the various SQL queries in our codebase. Each query is also fully self contained. This probably means there is a lot of code duplication. It's not always obvious which specific queries to modify when we want to adjust some in-app behavior.

One approach would be to use views & materialized views to store some common intermediary tables that could be used for more straight forward analysis. 

One example is a daily user activity view. It would contain a row per unique day+user along with some summary statistics. Like number of answers, total BONK, etc. That existence of that view would make streak calculation more simple to write.

If we don't think carefully about which views are helpful, we would just wind up increasing our tech debt. In the worst case, we don't end up end up reusing any views between queries and just adding overhead of creating views via migrations.

marvinmarnold (approved):

yash431garg (approved):
LGTM

PR #1018, opened 2025-03-03
PROD-871: Fix 500 Deck Page Error

neel-ds:
Handle null previous deck in getNextDeckId query (home). Add a test case to handle this edge case.


yash431garg (approved):
LGTM

samott (approved):

PR #1020, opened 2025-03-03
PROD-846: Solve 400 error on application

neel-ds:
- Fix TypeError: t is not iterable [Sentry](https://gator-labs.sentry.io/issues/6290534313/?environment=vercel-production&environment=production&project=4507686429130752&query=is%3Aunresolved%20iterable&referrer=issue-stream&statsPeriod=7d&stream_index=0)
  - Solution: Add sanity checks in formattedData useMemo (PaidDeckFeed.tsx and FreeDeckFeed.tsx)
- Fix Invalid URL / InApp Router Problem [Sentry](https://gator-labs.sentry.io/issues/5660041045/?project=4507686429130752&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=0) and [issue](https://gator-labs.sentry.io/issues/6317462856/?environment=vercel-production&project=4507686429130752&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=24h&stream_index=0)
  - Solution: Handle error in `useInfiniteQuery#queryfn` (PaidDeckFeed.tsx and FreeDeckFeed.tsx)
- Fix warnings of HTML tag props (also appears in console log)
  - Solution: All icon components are updated as per warning
Screenshot 2025-03-03 at 3 42 29 PM

- Removed stale requests of legacy components in application/page.tsx and stale legacy components or queries
- Typo in error bg patten svg


neel-ds:
Couldn’t reproduce the 400 error in prod, staging, or locally. Based on Sentry logs, it seems fixed, but it needs a thorough review to confirm the fix actually resolves the issue. One speculation was session expiration, but that seems unlikely since the JWT expiration is 7 days.

samott:
Is this still a draft @neel-ds? The changes make sense to me.

neel-ds:
> Is this still a draft @neel-ds? The changes make sense to me.

Yes, need to follow up on one unsolved potential issue behind 400.

neel-ds:
Error will be handled like below simulation: 

https://github.com/user-attachments/assets/08efa184-20e4-44e4-878c-7b4caf499f8d

samott (approved):
Looks good to me. I like the refetch feature.

app/components/FreeDeckFeed/FreeDeckFeed.tsx
```diff
@@ -1,35 +1,62 @@
...
+  const { data, fetchNextPage, hasNextPage, isFetching, refetch } =
+    useInfiniteQuery({
+      queryKey: ["free-decks"],
+      queryFn: ({ pageParam }) => {
+        try {
+          if (hasError) setHasError(false);
+          return getFreeDecks({ pageParam });
+        } catch {
+          setHasError(true);
+          return []; // Return empty array to prevent cascading error
```
marvinmarnold, line 22: Will this result in all of the data from the page being erased when it hits this point? Or will existing data persist?

marvinmarnold: Oh based on https://github.com/gator-labs/chomp/pull/1020#issuecomment-2708160938 I do think it clears all data on failure. Is it possible to swallow the error and keep the existing data on screen?

neel-ds: Correct, it will clear all data on failure. I was considering an edge case related to Replay—if a user opens a deck from the deck carousel, completes it, and then returns home, an error at that moment could result in invalidated cached data. If the error is swallowed, the answered deck might still appear incorrectly.

marvinmarnold: What would be the root cause of an error you are imagining that causes an error after answering a deck and returning home.

I'm thinking of the case where Cloudflare security check times out after leaving a tab open for a long time. Given these changes, I would expect users to always see an error in that case. I think this will happen a lot.

marvinmarnold: Let's move forward with reloading page when error occurs.

samott (approved):

yash431garg:
@neel-ds have you looked in the history page? Also are still working on this PR?

neel-ds:
> @neel-ds have you looked in the history page? Also are still working on this PR?

This is specifically for /application route. It's completed and need testing env to simulate behaviour.

yash431garg:
@neel-ds what are the next step here? Should we review again or you are still working on this PR?

neel-ds:
> @neel-ds what are the next step here? Should we review again or you are still working on this PR?

It's ready for review. Take notes from above simulated test env.

neel-ds:
I removed the simulated error environment. It's ready for merge.

PR #1021, opened 2025-03-04
fix: update open mystery box hub action

yash431garg:


app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -132,126 +131,136 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
...
-              },
-              data: {
-                status: EMysteryBoxStatus.Unopened,
+    try {
+      await prisma.$transaction(
+        async (tx) => {
+          await tx.mysteryBox.updateMany({
+            where: {
+              id: {
+                in: mysteryBoxIds,
```
neel-ds, line 223: it's better to rollback only for reward list ids not args based mystery box ids.

app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -132,126 +131,136 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
...
+      );
+    } catch (error) {
+      const openMysteryBoxHubError = new OpenMysteryBoxHubError(
+        `Trouble rolling back data with User id: ${payload.sub} (wallet: ${userWallet.address}) for Mystery Boxes: ${mysteryBoxIds}`,
+        { cause: error },
+      );
+      Sentry.captureException(openMysteryBoxHubError, {
+        extra: {
+          mysteryBoxIds,
+          userId: payload.sub,
```
neel-ds, line 256: can we have extra data of reward list mysterybox IDs?

app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -122,7 +121,7 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
   );
 
   const bonkAddress = process.env.NEXT_PUBLIC_BONK_ADDRESS || "";
-
+  const txHash: string | null = null;
```
neel-ds, line 124: What's the use of null constant?

__tests__/actions/open-mystery-box-hub.test.ts
```diff
@@ -0,0 +1,202 @@
...
+        },
+      ],
+    });
+
+    mysteryBoxId = mb.id;
+    mysteryBoxId1 = mb1.id;
+  });
+
+  afterAll(async () => {
+    await prisma.wallet.deleteMany({
```
neel-ds, line 133: Sequence of deleting data should align with data dependency:

1. FATL
2. MBP
3. MBT
4. MB
5. Wallet
6. User

neel-ds (changes requested):

__tests__/actions/open-mystery-box-hub.test.ts
yash431garg: Seems to a enhancement can be ignored for now.

app/actions/mysteryBox/openMysteryBoxHub.ts
yash431garg: It can be null as well.

app/actions/mysteryBox/openMysteryBoxHub.ts
yash431garg: Sure added

app/actions/mysteryBox/openMysteryBoxHub.ts
yash431garg: Thanks, I have added this

__tests__/actions/open-mystery-box-hub.test.ts
neel-ds: If you follow your code path manually, it won't allow you to delete because of dependency constraint. I feel this should be included in this PR so that it doesn't pile up corrupt data in DB.

app/actions/mysteryBox/openMysteryBoxHub.ts
neel-ds: I think it should be "let" to declare a variable whose state can be changed (for e.g on L126)

__tests__/actions/open-mystery-box-hub.test.ts
yash431garg: It won't fail. I have tested.

app/actions/mysteryBox/openMysteryBoxHub.ts
yash431garg: We are declaring the const and later assigning it a value. So, It shouldn't be an issue.

PR #1023, opened 2025-03-04
PROD-876: Update Deck Card Stack Page

neel-ds:
Note: This design will show No rewards for all legacy decks in Daily Deck Stack

Screenshot 2025-03-05 at 1 04 19 AM
Screenshot 2025-03-05 at 1 05 00 AM
Screenshot 2025-03-05 at 12 55 48 AM


app/stacks/[id]/page.tsx
```diff
@@ -62,40 +62,42 @@ const StackPage = async ({ params: { id } }: PageProps) => {
...
-            activeFromDate={deck.activeFromDate || deck.createdAt}
-            userId={user?.id}
-          />
-        ))}
+        {stack.deck.map((deck) => {
+          const totalCreditCost = deck.deckQuestions.reduce((total, dq) => {
+            return total + (dq.question.creditCostPerQuestion || 0);
+          }, 0);
+
+          const totalRewardAmount = deck.deckQuestions.reduce((total, dq) => {
```
Comment (yash431garg) [inline app/stacks/[id]/page.tsx:70]: Can we make a utility for these as we use this functionality in multiple places?

app/queries/stack.ts
```diff
@@ -185,11 +186,6 @@ export async function getAllStacks() {
             include: {
               question: {
                 include: {
-                  chompResults: {
-                    where: {
```
yash431garg, line 189: Won't it affect the old decks?

app/queries/stack.ts
```diff
@@ -236,16 +231,11 @@ export async function getDailyDecks() {
                 include: {
                   questionAnswers: {
                     where: {
-                      userId,
+                      userId: userId ?? "00000000-0000-0000-0000-000000000000",
```
yash431garg, line 234: I didn't understand why we have to do this.

app/queries/stack.ts
```diff
@@ -93,56 +78,72 @@ export async function getStack(id: number) {
...
-      // Group 3: Closed answer periods (descending `revealAtDate`)
-      if (isClosedA && isClosedB) {
-        return revealAtDateB.getTime() - revealAtDateA.getTime();
+      return {
+        ...deck,
+        totalCreditCost,
+        totalRewardAmount,
+      };
+    })
+    .sort((a, b) => {
```
yash431garg, line 97: I didn't understand this change

app/queries/stack.ts
```diff
@@ -54,24 +53,10 @@ export async function getStack(id: number) {
             include: {
               question: {
                 include: {
-                  chompResults: {
-                    include: {
-                      question: true,
```
yash431garg, line 59: I am not confident that should we remove chomp result or not

yash431garg (changes requested):
I have tested the changes for the ticket those changes looks good. But I am not sure about removing the whole reveal logic from stacks.

app/stacks/[id]/page.tsx
Comment (neel-ds) [reply app/stacks/[id]/page.tsx]: Good catch, removed repeating code.

app/queries/stack.ts
neel-ds: If user isn't logged in, match against a non-existent user to avoid pulling records for ALL users.

app/queries/stack.ts
neel-ds: This sort the decks based on activeFromDate and reveal date.

app/queries/stack.ts
neel-ds: Yes, I'm aware. This will be updated based on Product decision.

gatorthana (approved):

app/queries/stack.ts
neel-ds: Product said that mark legacy deck as practice deck. So, it's fine to remove the chompResult.

app/queries/stack.ts
neel-ds: We are marking legacy decks as Practice deck. It's fine to remove chompresult as we support credit now.

neel-ds:
> I have tested the changes for the ticket those changes looks good. But I am not sure about removing the whole reveal logic from stacks.

Yihwan confirmed to show legacy decks as practice decks. So, it doesn't require chompResult in query.

marvinmarnold:
1. "Decks" cut off
Screenshot 2025-03-08 at 16 08 19

2. "I am not sure about removing the whole reveal logic from stacks."
I'm also confused by this. Why was this removed. I created a paid deck in a stack and it no longer shows a button linking to reveal / go to mystery box page. Similarly, it's weird that I can't even enter the stack if I've already answered all the decks in the stack.

https://github.com/user-attachments/assets/f4ea95ce-090d-4361-946c-b448d36f4f15

neel-ds:
> 1. "Decks" cut off
>    Screenshot 2025-03-08 at 16 08 19
> 2. "I am not sure about removing the whole reveal logic from stacks."
>    I'm also confused by this. Why was this removed. I created a paid deck in a stack and it no longer shows a button linking to reveal / go to mystery box page. Similarly, it's weird that I can't even enter the stack if I've already answered all the decks in the stack.
> 

Previously, the reveal logic was based on isAfter(revealAtDate) and the existence of records in chompResult. With the new reveal flow, it should be determined using different criteria. I’ve asked Yihwan for clarification on that. Good catch, I will fix the disabled stack entry.

neel-ds:
Stack card state is fixed. Ready for final review! @samott @yash431garg

marvinmarnold:
Can you look into what's going wrong here?
https://github.com/user-attachments/assets/67857be7-8f69-4e3c-9fb5-5a809f143402

neel-ds:
> Can you look into what's going wrong here? https://github.com/user-attachments/assets/67857be7-8f69-4e3c-9fb5-5a809f143402

### Source: Vercel Logs 
[Error: Input buffer contains unsupported image format]

I think it was due to broken image and as soon as I changed the image, it's working now. I will add image placeholder so incase if it's broken then can handle it.

samott (approved):

marvinmarnold (approved):

PR #1033, opened 2025-03-06
PROD-753: Bulk purchase credits in purchase credits modal

samott:
This implements the credit pack options in the bulk buy modal.

Credit packs are stored in the database. They store the current credit cost as well as the discounted rate, so can be referenced in the FATL to see historical prices (if the credit cost ever changes). That does mean the existing records should be immutable though (a possible weakness of this design... easy to forget). The current prisma schema uses UUIDs for the packs, but this could be changed to a descriptive const-style string.

Unfortunately. I had to duplicate the `creditPack` reference in `ChainTx` as the process-credits cronjob needs it to recover any failed transactions. It's not ideal, but keeping the pack info only in `ChainTx` also doesn't feel right. I'm open to alternative ideas.

The credits logic is a bit complicated, so here is a reminder of the flow:

- Drawer component:
  - [FE] processCreditPurchase()
    - [FE] createCreditPurchaseTransaction()
    - [SERVER ACTION] initiateCreditPurchase()
      - [BE] createSignedSignatureChainTx()
      - [BE] processTransaction()
        - [BE] verifyPayment()
        - [BE] updateTxStatusToConfirmed()

creditpacks

### DIFF
--- a/__tests__/actions/credits/updateTxStatusConfirm.test.ts
+++ b/__tests__/actions/credits/updateTxStatusConfirm.test.ts
@@ -3,6 +3,7 @@ import prisma from "@/app/services/prisma";
 import { updateTxStatusToConfirmed } from "@/lib/credits/updateTxStatusConfirm";
 import { generateUsers } from "@/scripts/utils";
 import { faker } from "@faker-js/faker";
+import { CreditPack } from "@prisma/client";
 import {
   EChainTxStatus,
   EChainTxType,
@@ -17,7 +18,11 @@ jest.mock("p-retry", () => jest.fn().mockImplementation((fn) => fn()));
 
 describe("updateTxStatusToConfirmed", () => {
   const CREDIT_CONFIRM_SIGNATURE = "CREDIT_CONFIRM_SIGNATURE_" + "X".repeat(62);
+  const CREDIT_CONFIRM_SIGNATURE_2 =
+    "CREDIT_CONFIRM_SIGNATURE_" + "Y".repeat(62);
   let user: { id: string; wallet: string };
+
+  let creditPack: CreditPack;
 
   beforeAll(async () => {
     const users = await generateUsers(1);
@@ -39,6 +44,15 @@ describe("updateTxStatusToConfirmed", () => {
       },
     });
 
+    creditPack = await prisma.creditPack.create({
+      data: {
+        amount: 100,
+        costPerCredit: "0.00002",
+        originalCostPerCredit: "0.001",
+        isActive: true,
+      },
+    });
+
     await prisma.chainTx.create({
       data: {
         hash: CREDIT_CONFIRM_SIGNATURE,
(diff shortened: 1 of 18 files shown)


marvinmarnold:
Let's both try to remember to talk about this at the end of the next standup. I don't have any concerns but would just like to talk through it.

marvinmarnold (changes requested):
I don't see packs in preview link 

Screenshot 2025-03-10 at 12 39 42

samott:
> I don't see packs in preview link

I just configured some packs in the database. Please try again.

marvinmarnold:
Screenshot 2025-03-10 at 19 16 57

Is there a chance you messed up the DB entries? Or maybe you are forgetting to multiply the unit price by the number of units. Whatever is going on, it gets more expensive for fewer units.

samott:
> Is there a chance you messed up the DB entries? Or maybe you are forgetting to multiply the unit price by the number of units. Whatever is going on, it gets more expensive for fewer units.

The unit price is shown for each item. I figured we are already showing the total at the bottom.

But I can update it to show the total for each pack.

marvinmarnold:
Looks like you were totally right the first time, my bad. After this gets merged to staging, could you setup the packs so they follow the discount pattern below? The smallest two boxes should not show any discount.
```
quantities: 500 for 0.125 SOL (discounted from 0.25 SOL), 250 for 0.125 SOL (no discount), 100 for 0.05 SOL (no discount)
````

I honestly just reviewed AC for first time and I don't think this behavior is implemented.
```
user can manually input. -/+ buttons increment by 1 (integer), min 0, no max. 
updating manual input through text input or -/+ buttons deselects any selected credit pack
can user manually input decimal? defer to dev on this. in any case, +/- button increments by integer only. 
```

What were your thoughts about this tracking? I think you are planning to make all this info available from DB? @Yihwan were you expecting in Mixpanel?
```
tracking: 
admin user should be able to track whether a single credit purchase transaction was from a pack. this data in turn could be aggregated (counted) to see how many packs were purchased in total. 
for now we can use constants (no need to track discounts separately in detail for now):
500-credits_0.125-SOL_discounted-from-0.25-SOL
250-credits_0.125-SOL
100-credits_0.05-SOL
```

samott:
> Looks like you were totally right the first time, my bad. After this gets merged to staging, could you setup the packs so they follow the discount pattern below? The smallest two boxes should not show any discount.
> 

Ok, I'll revert that change.

> ```
> user can manually input. -/+ buttons increment by 1 (integer), min 0, no max. 
> updating manual input through text input or -/+ buttons deselects any selected credit pack
> can user manually input decimal? defer to dev on this. in any case, +/- button increments by integer only. 
> ```

I'll update the drawer to keep the buttons enabled.

> What were your thoughts about this tracking? I think you are planning to make all this info available from DB? @Yihwan were you expecting in Mixpanel?
> 
> ```
> tracking: 
> admin user should be able to track whether a single credit purchase transaction was from a pack. this data in turn could be aggregated (counted) to see how many packs were purchased in total. 
> for now we can use constants (no need to track discounts separately in detail for now):
> 500-credits_0.125-SOL_discounted-from-0.25-SOL
> 250-credits_0.125-SOL
> 100-credits_0.05-SOL
> ```

I could certainly add the tracking in Mixpanel if needed. Seems to me the DB is the best way to query/analyse the data as we have everything there.

Yihwan:
hi! for trackig i'm hoping we can track this in db so we can query using retool.

marvinmarnold (approved):

neel-ds (approved):
LGTM

PR #1034, opened 2025-03-07
PROD-877: Add duplicate deck functionality to admin panel

samott:
- Add duplicated deck functionality to admin panel.
- Fix `deleteDeck()` function to remove `questionTags`.

### DIFF
--- a/app/actions/deck/deck.ts
+++ b/app/actions/deck/deck.ts
@@ -156,6 +156,14 @@ export async function deleteDeck(deckId: number) {
 
       await tx.deck.delete({ where: { id: deckId } });
 
+      await tx.questionTag.deleteMany({
+        where: {
+          questionId: {
+            in: questionIds,
+          },
+        },
+      });
+
       await tx.question.deleteMany({
         where: {
           id: {
--- a/app/actions/deck/deck.ts
+++ b/app/actions/deck/deck.ts
@@ -170,6 +170,93 @@ export async function deleteDeck(deckId: number) {
   revalidatePath("/admin/decks");
 }
 
+export async function copyDeck(deckId: number): Promise<number> {
+  const isAdmin = await getIsUserAdmin();
+
+  if (!isAdmin) {
+    redirect("/application");
+  }
+  const deck = await prisma.deck.findUnique({
+    where: { id: deckId },
+  });
+
+  if (!deck) throw new Error("Deck not found");
+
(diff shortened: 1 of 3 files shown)


neel-ds (approved):
LGTM! Can you move this action outside the app directory and rename it to something more specific like copyDeck.ts?

app/actions/deck/deck.ts
```diff
@@ -170,6 +178,100 @@ export async function deleteDeck(deckId: number) {
...
+    },
+  });
+
+  let newDeckId: number | undefined = undefined;
+
+  await prisma.$transaction(async (tx) => {
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    const { id, ...newDeck } = deck;
+
+    newDeck.deck = "Copy of " + newDeck.deck;
```
marvinmarnold, line 211: I think a few deck properties are missing.
Screenshot 2025-03-08 at 16 50 27

app/actions/deck/deck.ts
```diff
@@ -170,6 +178,100 @@ export async function deleteDeck(deckId: number) {
...
+    newDeck.activeFromDate = null;
+    newDeck.revealAtDate = null;
+    newDeck.createdAt = new Date();
```
marvinmarnold, line 214: Do these really need to be set explicitly, the schema uses default values.
Screenshot 2025-03-08 at 16 52 02

It's kind of a nit, but also @psypersky has been digging into potentially problematic ways of handling timezones that we've implemented.

app/actions/deck/deck.ts
```diff
@@ -170,6 +178,100 @@ export async function deleteDeck(deckId: number) {
...
+
+  if (!deck) throw new Error("Deck not found");
+
+  const deckQuestions = await prisma.deckQuestion.findMany({
+    where: { deckId },
+    include: {
+      question: {
+        include: {
+          questionOptions: true,
+          questionTags: true,
```
marvinmarnold, line 199: I guess this is fine, but we aren't using QuestionTags currently.

marvinmarnold (changes requested):

app/actions/deck/deck.ts
samott: Fixed to let Prisma set the dates.

app/actions/deck/deck.ts
samott: Which ones? They're all copied over I believe.

PR #1036, opened 2025-03-07
PROD-900: implement bot detector cronjob test

samott:
Fixes the bot detector which was, in fact, not working. Adds tests to prove it now works.


neel-ds (approved):
LGTM!

marvinmarnold (approved):
Can you elaborate on what wasn't working before this commit? Was it just the window that wasn't getting saved, or was threatLevel not being recorded at all? Will we need to go in and backfill anything?

PR #1037, opened 2025-03-08
PROD-888: complete campaign mystery box integration

samott:
The box type sent to the mechanism engine depends on the campaign week.

### DIFF
--- a/prisma/migrations/20250308130515_add_boxtype_to_campaign_mystery_box/migration.sql
+++ b/prisma/migrations/20250308130515_add_boxtype_to_campaign_mystery_box/migration.sql
@@ -1,0 +1,8 @@
+/*
+  Warnings:
+
+  - Added the required column `boxType` to the `CampaignMysteryBox` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- AlterTable
+ALTER TABLE "CampaignMysteryBox" ADD COLUMN     "boxType" TEXT NOT NULL;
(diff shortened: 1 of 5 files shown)


prisma/migrations/20250308130515_add_boxtype_to_campaign_mystery_box/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- AlterTable
+ALTER TABLE "CampaignMysteryBox" ADD COLUMN     "boxType" TEXT;
+
+UPDATE "CampaignMysteryBox" SET "boxType" = 'Bis1' WHERE "boxType" IS NULL;
```
marvinmarnold, line 4: @neel-ds you are doing this week's release. You will need to make sure that we haven't created a CampaignMysteryBox for week 2 by the time you release. Otherwise, it's boxType will be incorrectly set.

marvinmarnold (approved):

yash431garg:
@samott I think we need to clarify what happen if someone haven't claimed from the previous week box. We need to configure it in a way that user get the reward based on which week campaign the user is opening.

samott:
> @samott to modify script that creates campaign to include boxType

Anyone know where this script actually lives? Doesn't seem to be in 'scripts'. Cc: @yash431garg

psypersky (approved):
LGTM

PR #1038, opened 2025-03-10
PROD-884: Store signature in send bonk

neel-ds:
We store signature in credit payment even if it fails, this PR add similar update in claim MB. It will create ChainTx even before confirming tx. Later, update the status in client.

### DIFF
--- a/app/actions/mysteryBox/openMysteryBoxHub.ts
+++ b/app/actions/mysteryBox/openMysteryBoxHub.ts
@@ -7,7 +7,6 @@ import {
   EBoxPrizeStatus,
   EBoxPrizeType,
   EChainTxStatus,
-  EChainTxType,
   EMysteryBoxStatus,
   FungibleAsset,
   TransactionLogType,
@@ -120,7 +119,6 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
     0,
   );
 
-  const bonkAddress = process.env.NEXT_PUBLIC_BONK_ADDRESS || "";
   let txHash: string | null = null;
   try {
     if (totalBonkAmount > 0) {
@@ -142,17 +140,13 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
             throw new Error("Treasury address not defined");
           }
 
-          await tx.chainTx.create({
+          await tx.chainTx.update({
             data: {
-              hash: txHash,
-              wallet: treasury,
-              recipientAddress: userWallet.address,
-              type: EChainTxType.MysteryBoxClaim,
               status: EChainTxStatus.Finalized,
-              solAmount: "0",
-              tokenAmount: totalBonkAmount.toString(),
-              tokenAddress: bonkAddress,
               finalizedAt: date,
+            },
+            where: {
+              hash: txHash,
             },
           });
         }
--- a/app/utils/sendBonk.ts
+++ b/app/utils/sendBonk.ts
@@ -2,6 +2,8 @@ import {
   TransactionFailedError,
   TransactionFailedToConfirmError,
 } from "@/lib/error";
+import { EChainTxType } from "@prisma/client";
+import { EChainTxStatus } from "@prisma/client";
 import * as Sentry from "@sentry/nextjs";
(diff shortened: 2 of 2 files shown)


app/utils/sendBonk.ts
```diff
@@ -159,9 +162,22 @@ export const sendBonk = async (toWallet: PublicKey, amount: number) => {
...
     });
+
+    await prisma.chainTx.create({
+      data: {
+        hash: signature,
+        status: EChainTxStatus.New,
+        solAmount: "0",
+        wallet: fromWallet.publicKey.toBase58(),
+        recipientAddress: toWallet.toBase58(),
+        type: EChainTxType.MysteryBoxClaim,
```
samott, line 173: This now ties `sendBonk()` to this particular TX type... which is not ideal.

neel-ds: We only use sendBonk for MysteryBox. Do you think we should generalize this utility?

psypersky: Hey. I still don't know much but it does sounds like a good idea to send the type as a param or changing the name of the function to sendBonkForMysteryBoxClaim so new people (like me) immediately know that :)

psypersky (approved):

app/actions/mysteryBox/openMysteryBoxHub.ts
```diff
@@ -142,18 +140,14 @@ export const openMysteryBoxHub = async (mysteryBoxIds: string[]) => {
             throw new Error("Treasury address not defined");
           }
 
-          await tx.chainTx.create({
+          await tx.chainTx.update({
```
marvinmarnold, line 143: A little bit of a nit, but I think this update to `Finalized` would make more sense at the end of sendBonk. As far as I can tell, there's no other relevant confirmation code between sendBonk, and when tx.chainTx.update is invoked.

So it makes more sense to create and update the ChainTx fully contained within sendBonk.

marvinmarnold (changes requested):

app/utils/sendBonk.ts
neel-ds: Understood. I will add type param.

app/actions/mysteryBox/openMysteryBoxHub.ts
neel-ds: Understood, will keep the entire flow in sendBonk utility

yash431garg (changes requested):

app/utils/sendBonk.ts
```diff
@@ -186,7 +213,14 @@ export const sendBonk = async (toWallet: PublicKey, amount: number) => {
     if (!signature) {
       throw new Error("Transaction signature is null");
     }
-    await checkTransactionStatus(signature);
+    const isConfirmed = await checkTransactionStatus(signature);
+
+    if (isConfirmed) {
+      await prisma.chainTx.update({
```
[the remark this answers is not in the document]
neel-ds: It is suggested to move status to sendBonk by Marvin as there is no confirmation code between sendBonk and openMysteryBoxHub.

yash431garg (approved):

psypersky (approved):
LGTM but keep in mind in new :p

app/utils/sendBonk.ts
```diff
@@ -152,6 +166,19 @@ export const sendBonk = async (toWallet: PublicKey, amount: number) => {
   const versionedTransaction = new VersionedTransaction(v0message);
   versionedTransaction.sign([fromWallet]);
 
+  await prisma.chainTx.create({
+    data: {
+      hash: base58.encode(versionedTransaction.signatures[0]),
```
marvinmarnold, line 171: It just occurred to me that this might be wrong if the tx has multiple instructions. This may happen if ATA is created for transfer also.

`base58.encode(versionedTransaction.signatures)` may work but I'm not sure.

Can you please confirm behavior if ATA is created as part of tx as well?

neel-ds: It's same outcome for transfer with ATA.

PR #1043, opened 2025-03-12
PROD-904: Show Decks on History Tab

neel-ds:
Currently, it shows each question answered by user. This adjust will only show (revealed and answered) deck by user. 

Screenshot 2025-03-12 at 6 11 21 PM


app/_reactQuery/useInfiniteHistoryDecks.ts
```diff
@@ -0,0 +1,65 @@
...
+    isFetchingNextPage,
+    ...rest
+  } = useInfiniteQuery({
+    queryKey: ["history-decks"],
+    queryFn: ({ pageParam }) => getHistoryDecks({ pageParam }),
+    initialPageParam: 1,
+    getNextPageParam: (lastPage, allPages) => {
+      return lastPage.length ? allPages.length + 1 : undefined;
+    },
+    staleTime: Infinity,
```
marvinmarnold, line 24: Will this cause the data to never refresh?

app/_reactQuery/useInfiniteHistoryDecks.ts
```diff
@@ -0,0 +1,65 @@
+import { useInfiniteQuery } from "@tanstack/react-query";
+import { useCallback, useMemo, useRef } from "react";
+
+import { getHistoryDecks } from "../actions/history";
+
+export default function useInfiniteHistoryDecks() {
+  const observer = useRef<IntersectionObserver>();
```
marvinmarnold, line 7: Interesting pattern. Is there some reference you got this from? It looks like this is for infinite query. I think @Yihwan prefers to have a load more button and that's what we are doing elsewhere in the app.

app/queries/history.ts
```diff
@@ -362,3 +363,51 @@ WHERE
...
 }
+
+export async function getAnsweredDecksForHistory(
+  userId: string,
+  pageSize: number,
+  currentPage: number,
+): Promise<DeckHistoryItem[]> {
+  const offset = (currentPage - 1) * pageSize;
+
+  const result: DeckHistoryItem[] = await prisma.$queryRaw`
```
marvinmarnold, line 374: I'm surprised there's no other query for getting answered decks.

Yihwan (approved):

app/queries/history.ts
yash431garg: I guess it is possible because we don't have any requirement for show answered deck earlier.

app/queries/history.ts
```diff
@@ -362,3 +363,51 @@ WHERE
...
+      AND d."revealAtDate" <= NOW()
+      AND EXISTS (
+        SELECT 1
+        FROM public."DeckQuestion" dq
+        JOIN public."Question" q ON dq."questionId" = q.id
+        JOIN public."QuestionOption" qo ON qo."questionId" = q.id
+        JOIN public."QuestionAnswer" qa ON qa."questionOptionId" = qo.id
+        WHERE dq."deckId" = d.id
+        AND qa."userId" = ${userId}
+        AND qa."status" IN ('Submitted', 'Viewed')
```
yash431garg, line 403: I think we should also include Skipped.

components/HistoryNew/HistoryDeckCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { getDeckPath } from "@/lib/urls";
+import { DeckHistoryItem } from "@/types/history";
+import { TrophyIcon } from "lucide-react";
+import Image from "next/image";
+
+interface HistoryDeckCardProps {
+  deck: DeckHistoryItem;
+}
+
+export const HistoryDeckCard = ({ deck }: HistoryDeckCardProps) => {
```
yash431garg, line 17: Can't we reuse the similar card which we show in stack or application route?

app/_reactQuery/useInfiniteHistoryDecks.ts
neel-ds: This was used in previous implementation of history tab (useInfiniteQuestionsHistory). Ok, I will replace it with load more.

app/_reactQuery/useInfiniteHistoryDecks.ts
neel-ds: Yes, this was used in previous implementation of history tab (useInfiniteQuestionsHistory).

neel-ds: I removed it

app/queries/history.ts
neel-ds: We only have one query - getDecksHistory (for legacy decks). Thus, created new one for credit flow

components/HistoryNew/HistoryDeckCard.tsx
neel-ds: While the cards share visual similarities but they have fundamentally different logic (stack deck has login feat), state (btn state with timer), and interaction patterns (progress bar). Unifying them would create unnecessary complexity and maintenance challenges that outweigh the benefits of code reuse.

marvinmarnold (approved):

PR #1047, opened 2025-03-13
fix: stack deck state

neel-ds:
Fix the stack deck state when question is skipped

### DIFF
--- a/app/queries/stack.ts
+++ b/app/queries/stack.ts
@@ -98,11 +98,7 @@ export async function getStack(id: number) {
         deck.deckQuestions
           .filter((dq) =>
             dq.question.questionOptions.some((qo) =>
-              qo.questionAnswers.some(
-                (qa) =>
-                  qa.userId === userId &&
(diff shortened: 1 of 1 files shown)


app/queries/stack.ts
```diff
@@ -98,11 +98,7 @@ export async function getStack(id: number) {
         deck.deckQuestions
           .filter((dq) =>
             dq.question.questionOptions.some((qo) =>
-              qo.questionAnswers.some(
-                (qa) =>
-                  qa.userId === userId &&
-                  (qa.status === "Submitted" || qa.status === "Viewed"),
-              ),
+              qo.questionAnswers.some((qa) => qa.userId === userId),
```
marvinmarnold, line 101: Ideally all bug fixes have tests

marvinmarnold (approved):

app/queries/stack.ts
neel-ds: Will follow up on this after the release

PR #1048, opened 2025-03-16
PROD-901: Display 2nd order average based on selected responses

samott:
Resolves an issue with calculated second order percentages by limiting denominator to selected answers rather than entire cohort.

The problem is these incorrect results are stored in the database and used elsewhere; I have created follow-up tickets to look into this.

### DIFF
--- a/actions/answers/answerQuestion.ts
+++ b/actions/answers/answerQuestion.ts
@@ -16,7 +16,6 @@ import {
 } from "@prisma/client";
 import * as Sentry from "@sentry/nextjs";
 import { revalidatePath } from "next/cache";
-import { release } from "os";
 
 export type SaveQuestionRequest = {
   questionId: number;
@@ -191,7 +190,6 @@ export async function answerQuestion(request: SaveQuestionRequest) {
     );
     Sentry.captureException(answerError);
     await Sentry.flush(SENTRY_FLUSH_WAIT);
-    release();
     throw error;
   }
 }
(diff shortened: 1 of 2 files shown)


yash431garg (approved):

actions/answers/answerQuestion.ts
```diff
@@ -191,7 +190,6 @@ export async function answerQuestion(request: SaveQuestionRequest) {
     );
     Sentry.captureException(answerError);
     await Sentry.flush(SENTRY_FLUSH_WAIT);
-    release();
```
marvinmarnold, line 194: This is weird. Any idea what the intention was? This fix is unrelated to the PR?

app/queries/answerPercentageQuery.ts
```diff
@@ -41,7 +41,7 @@ export async function answerPercentageQuery(questionOptionIds: number[]) {
                   select round(avg(percentage))
                   from public."QuestionAnswer"
                   join public."User" u on "userId" = u."id"
-                  where "questionOptionId" = qo."id" and "status" = 'Submitted' and (u."threatLevel" IS NULL OR u."threatLevel" IN (${EThreatLevelType.ManualAllow}, ${EThreatLevelType.PermanentAllow}))
+                  where "questionOptionId" = qo."id" and "selected" is true AND "status" = 'Submitted' and (u."threatLevel" IS NULL OR u."threatLevel" IN (${EThreatLevelType.ManualAllow}, ${EThreatLevelType.PermanentAllow}))
```
marvinmarnold, line 44: Normally, I'd be more insistent about a unit test to go along with this. But we are going to change all of this soon: https://linear.app/gator/issue/PROD-922/make-mechanism-engine-responsible-for-calculations

marvinmarnold (approved):

actions/answers/answerQuestion.ts
marvinmarnold: Ah ok.

PR #1049, opened 2025-03-17
Prod 910: Rate limit bonk sent

yash431garg:
Description: Add rate limit on bonk transfer from treasury with in an hourly window. Implemented using sliding window algorithm. 

Test: Transaction should not be submitted if the window is out of limit.


yash431garg:
I am still testing it in preview.

lib/mysteryBox.ts
```diff
@@ -45,15 +46,32 @@ export async function sendBonkFromTreasury(
...
+    } else {
+      const bonkRateLimitExceedError = new BonkRateLimitExceedError(
+        `User with id: ${userId} (wallet: ${address}) is having trouble claiming. `,
+      );
+      Sentry.captureException(bonkRateLimitExceedError, {
+        extra: {
+          userId,
+          walletAddress: address,
+          rewardAmount,
+          remainingWindowAmount: 0,
```
samott, line 70: Is `remainingWindowAmount` telling us anything here?

lib/mysteryBox.ts
```diff
@@ -45,15 +46,32 @@ export async function sendBonkFromTreasury(
...
+    const isWithinBonkHourlyLimit = await checkBonkRateLimit(rewardAmount);
+
+    if (isWithinBonkHourlyLimit) {
+      const sendTx = await sendBonk(
+        new PublicKey(address),
+        Math.round(rewardAmount * 10 ** 5),
+        type,
+      );
+      return sendTx;
+    } else {
```
samott, line 61: Minor style nit here, but it's generally better to have the exception case in an if statement first and then the main case doesn't need to be nested.

marvinmarnold (changes requested):
checkBonkRateLimit also needs to be checked in sendBonk or we won't catch bots that skip the UI.

lib/mysteryBox.ts
yash431garg: Yes, I forgot to push the updated code.

yash431garg:
@marvinmarnold on your point  on using the checkBonk utility in sendBonk. We don't need to do that as sendBonk is a server-only component.

lib/mysteryBox.ts
```diff
@@ -45,15 +46,32 @@ export async function sendBonkFromTreasury(
...
+      const bonkRateLimitExceedError = new BonkRateLimitExceedError(
+        `User with id: ${userId} (wallet: ${address}) is having trouble claiming. `,
```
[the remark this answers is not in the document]
yash431garg: Thank you, I have added this in the next commit.

lib/bonk/rateLimiter.ts
```diff
@@ -0,0 +1,28 @@
...
+
+import { kv } from "../kv";
+
+const HOURLY_LIMIT = Number(process.env.BONK_HOURLY_LIMIT) || 10000000;
+// Create a rate limiter that allows 10M tokens per hour
+export const bonkRateLimiter = new Ratelimit({
+  redis: kv,
+  limiter: Ratelimit.slidingWindow(HOURLY_LIMIT, "3600 s"),
+  analytics: true,
+  prefix: "bonk_distribution",
```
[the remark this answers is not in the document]
yash431garg: I have removed this prefix.

neel-ds (commented):
LGTM. How can I simulate test env?

yash431garg:
> LGTM. How can I simulate test env?

Ready for testing. The Bonk window is 25 right now per hour.

marvinmarnold:
Left a comment in Linear because attachment too big: https://linear.app/gator/issue/PROD-910/rate-limit-bonk-sent#comment-f3a1bb15

app/utils/sendBonk.ts
```diff
@@ -30,12 +30,14 @@ import { sleep } from "./sleep";
...
- * @returns The signature of the transaction or null if the transaction fails
- */
+
+Send Bonk to a wallet
+@warning 🚨 DO NOT USE THIS METHOD DIRECTLY - Use sendBonkFromTreasury instead which implements rate limiting
+@param toWallet - The recipient wallet address
+@param amount - The amount of Bonk to send
+@param type - The trigger type of the transaction
+@returns The signature of the transaction or null if the transaction fails
+*/
```
samott, line 40: nit: the comment formatting is weird now

lib/mysteryBox.ts
```diff
@@ -45,18 +46,38 @@ export async function sendBonkFromTreasury(
...
+        rewardAmount,
+        remainingWindowLimit: remainingLimit,
+      },
+    });
+    return null;
+  }
+  // Main case (successful path)
+  const sendTx = await sendBonk(
+    new PublicKey(address),
+    Math.round(rewardAmount * 10 ** 5),
```
samott, line 77: Probably not worth fixing this here but it kind of bugs me that we scale up the amount here then down again in for the DB insert in `sendBonk()` - it would probably be better to do the `* 10**5` inside `sendBonk()` since it's also not clear from the parameter name whether the value is supposed to be scaled up.

app/utils/sendBonk.ts
yash431garg: I was also confused. Where should I place the comment. Any suggestion?

lib/mysteryBox.ts
yash431garg: You are right. Actually It was like this previously and I just updated the code structure. But now I feel that we should update this.

yash431garg: It was a great catch Shaun. I have update the  code as a part of this PR. Please review that specifically. 

Related Commit: https://github.com/gator-labs/chomp/pull/1049/commits/daf32f4548bc2a3f1cfbfa5da7779f3e1ef0c382

app/utils/sendBonk.ts
samott: The location is fine, it's just the spacing is off.

samott (approved):
Looks good to me with the recent changes.

PR #1051, opened 2025-03-18
PROD-902: enable-logged-out-users-to-see-deckinfo-and-stacks-page

psypersky:

Show decks to not logged in users

To test make sure all deck related stuff still works ok and not logged in users can see decks

NOTE: We only show decks that are active and not revealed to logged out users


app/queries/deck.ts
```diff
@@ -81,9 +81,9 @@ export async function getDailyDeck() {
 
 export async function getDeckQuestionsForAnswerById(deckId: number) {
   const payload = await getJwtPayload();
-  if (!payload?.sub) return null;
+  const userId = payload?.sub || null;
 
-  const userId = payload.sub;
+  console.log("userId", userId);
```
marvinmarnold, line 86: Remove log

app/queries/deck.ts
```diff
@@ -96,11 +96,7 @@ export async function getDeckQuestionsForAnswerById(deckId: number) {
             include: {
               questionOptions: {
                 include: {
-                  questionAnswers: {
-                    where: {
-                      userId,
-                    },
-                  },
+                  questionAnswers: userId ? { where: { userId } } : true, // Include all if null
```
marvinmarnold, line 99: If a user is not logged in, I would think they should see none of the QuestionAnswers from other users. This allows them to see all of other users answers?

app/application/layout.tsx
```diff
@@ -21,6 +21,8 @@ export default async function Layout({ children }: PageLayoutProps) {
     getTransactionHistory(),
   ]);
 
+  const isUserLoggedIn = !!user?.id;
```
marvinmarnold, line 24: I recommend we do this via context/provider instead of prop drilling/passing. We're going to make more parts of the app accessible to logged out users and passing the param around will be messy. Current user is generally a useful thing to be able to access via context.

app/components/Navbar/Navbar.tsx
```diff
@@ -34,29 +37,42 @@ export function Navbar({
...
+    event.preventDefault();
+    redirectToMainDomain();
+  };
+
   return (
     <nav className="flex justify-between w-full py-3 items-center fixed top-0 left-1/2 -translate-x-1/2 px-4 bg-gray-900 z-10 max-w-lg">
-      <Link href="/application">
-        <ChompFlatIcon fill="#fff" />
+      <Link
+        href={isUserLoggedIn ? "/application" : "#"}
```
marvinmarnold, line 48: Why not `href={isUserLoggedIn ? "/application" : "https://chomp.games"}`? It's weird to put an `a` tag inside a Link like this.

app/components/AuthRedirect/AuthRedirect.tsx
```diff
@@ -7,7 +7,16 @@ export const AuthRedirect = async () => {
   const jwt = await getJwtPayload();
   const path = headers().get("x-path");
 
-  if (!jwt) {
+  // add any path logged off users should be able to see here
+  const pathExceptionsOfAuthRedirect = ["/decks"];
```
marvinmarnold, line 11: This is inadvertently making /admin/decks accessible without login.

https://chomp-dev-git-prod-902-gator-labs.vercel.app/admin/decks

marvinmarnold (changes requested):

yash431garg (changes requested):

app/queries/deck.ts
psypersky: done

app/queries/deck.ts
psypersky: created a new query specifically for logged out users

app/components/Navbar/Navbar.tsx
psypersky: done

app/application/decks/[id]/page.tsx
```diff
@@ -1,102 +1,159 @@
...
+  }
+
+  let deckIn; // deck for logged in users
+  let deckOut; // deck for logged out users
+  if (isUserLoggedIn) {
+    deckIn = await getDeckQuestionsForAnswerById(currentDeckId);
+  } else {
+    deckOut = await getDeckForLoggedOutUsers(currentDeckId);
+  }
+
```
Comment (samott) [inline app/application/decks/[id]/page.tsx:39]: This seems a bit convoluted. Wouldn't it be easier just to do

```
if (!isUserLoggedIn) {
  const deck = await getDeckForLoggedOutUsers(currentDeckId);
  return ...;
}

const deck = await getDeckQuestionsForAnswerById(currentDeckId);

// logged out stuff
```

Comment (psypersky) [reply app/application/decks/[id]/page.tsx:39]: Yeah it is, it was mostly because the types are different, I was kind of fighting with typescript, let me try again to do it with one variable

Comment (psypersky) [reply app/application/decks/[id]/page.tsx:39]: ok, I tried again but the thing is that all the logic on the first part of the function needs the deck so can't really move it under if (!isUserLoggedIn)

and on the other part if I use the same variable for both type of decks typescript will scream at me because the types of both deck variables are different, so still think this is the best solution :/

Comment (marvinmarnold) [reply app/application/decks/[id]/page.tsx:39]: What about creating two new components `DeckForLoggedInUser` and `DeckForLoggedOutUser` then within this page doing

```
if (isUserLoggedIn) return <DeckForLoggedInUser  ... />
return <DeckForLoggedOutUser ... />
```

app/components/AuthRedirect/AuthRedirect.tsx
```diff
@@ -7,7 +7,18 @@ export const AuthRedirect = async () => {
   const jwt = await getJwtPayload();
   const path = headers().get("x-path");
 
-  if (!jwt) {
+  // add any path logged off users should be able to see here
+  // NOTICE: be sure to add the '/application' part or you could
+  // remove auth from an admin route
+  const pathExceptionsOfAuthRedirect = ["/application/decks"];
+  let pathExemptOfAuthR = false;
```
marvinmarnold, line 14: nit: naming
- Using exception and exemptions interchangeably
- Inconsistently using `Redirect` or just `R`
- Boolean variables should start with `is`

One solution would be `exemptAuthRedirectPaths` and `isPathExemptFromAuthRedirect`

marvinmarnold (commented):
Missing tests.

Also, during retro we agreed to no longer add new server actions.

app/application/decks/[id]/page.tsx
Comment (psypersky) [reply app/application/decks/[id]/page.tsx]: I refactored it, please tell me if you like this version more. With this will be trivial to separate them later each in its own component and we load everything on the server page :)

app/components/AuthRedirect/AuthRedirect.tsx
psypersky: fixed

app/application/decks/[id]/page.tsx
Comment (samott) [reply app/application/decks/[id]/page.tsx]: Refactor looks better, thanks!

samott:
Am I supposed to be able to see this page while logged out?

https://chomp-dkqhkzhjb-gator-labs.vercel.app/application/decks/6073

I get a 404...

psypersky:
@samott we only show not revealed decks to not logged in users for the moment, you can see this one for example

https://chomp-dev-git-prod-902-gator-labs.vercel.app/application/decks/7455

lib/data_gen/deck.ts
marvinmarnold, line 1: Cool utility. But shouldn't this be in __tests__ directory?

app/application/decks/[id]/page.tsx
```diff
@@ -1,102 +1,142 @@
...
+          freeExpiringDeckId={freeExpiringDeckId?.id ?? null}
+          blurData={blurData?.base64}
+        />
+      </div>
+    );
+  }
+
+  // User is logged in
+
+  const hydratedDeck = await getDeckQuestionsForAnswerById(currentDeckId);
```
Comment (marvinmarnold) [inline app/application/decks/[id]/page.tsx:85]: nit: I don't love the name hydratedDeck. Not immediately obvious what the difference between a deck and hydratedDeck is.

app/queries/deck.ts
```diff
@@ -79,6 +89,103 @@ export async function getDailyDeck() {
   };
 }
 
+/**
+ * Gets a deck and calculates its cost and reward for all its questions
+ * used for logged out users, only get active decks that are not revealed yet
+ */
+export async function getDeckForLoggedOutUsers(deckId: number) {
```
marvinmarnold, line 96: Consider renaming to getActiveDeckForLoggedOutUsers

__tests__/queries/deck/getDeckForLoggedOutUsers.ts
marvinmarnold, line 1: Should test more scenarios. Like making sure that already revealed decks don't show up. And that it does return daily decks.

app/application/decks/[id]/page.tsx
Comment (samott) [reply app/application/decks/[id]/page.tsx]: On closer inspection, I still don't love this solution - I feel like you're trying too hard to avoid code duplication but are making things more complicated, and I'm not clear on why we have to query the deck table twice.

That said, I think this is probably ok for now.

lib/data_gen/deck.ts
psypersky: Anything that you put on the tests directory need to have a test case , so I had to take it out

app/application/decks/[id]/page.tsx
Comment (psypersky) [reply app/application/decks/[id]/page.tsx]: yeah I guess 10 lines of repeated code on both side doesn't hurt, next time I'm there I'll just do that

app/application/decks/[id]/page.tsx
Comment (psypersky) [reply app/application/decks/[id]/page.tsx:85]: According to chatgpt that's the most common term , what is nit btw ?

Common Terms for a "Full" Deck (Including Relations)
Hydrated Deck ✅

Best choice – Comes from the concept of "hydration" (filling an object with all its related data).

Example: "We fetch a hydrated deck including cards, answers, and creator info."


Populated Deck (Mongoose/NoSQL term, but sometimes used in Prisma)

Used in MongoDB/Mongoose when referencing related documents.

Example: "The populated deck contains all its cards."


Eager-Loaded Deck (Prisma/ORM terminology)

Refers to loading relations upfront (vs. lazy-loading).

Example: "We eager-load the deck with its questions."


Full Deck / Complete Deck (Informal)

Simple and clear, but less technical.

Example: "API returns a full deck with all nested cards."

app/queries/deck.ts
psypersky: done

__tests__/queries/deck/getDeckForLoggedOutUsers.ts
psypersky: added another three scenarios

lib/data_gen/deck.ts
marvinmarnold: moved to different thread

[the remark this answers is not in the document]
psypersky: ready

app/queries/stack.ts
```diff
@@ -170,12 +170,6 @@ export async function getStack(id: number) {
 }
 
 export async function getStackImage(id: number) {
-  const payload = await getJwtPayload();
```
yash431garg, line 173: Why you removed this?

app/queries/deck.ts
```diff
@@ -16,6 +16,16 @@ import { getJwtPayload } from "../actions/jwt";
 import prisma from "../services/prisma";
 import { getTotalNumberOfDeckQuestions } from "../utils/question";
 
+export async function getRawDeck(deckId: number) {
+  const deck = await prisma.deck.findUnique({
```
yash431garg, line 20: I think we should add auth check here.

psypersky: we need it for not logged users too

samott (approved):

PR #1057, opened 2025-03-20
PROD-917: Remove lag from mystery box button state.

samott:
Ensure that the button in the mystery box hub can't be clicked again after a claim has completed (or during the claim process).


psypersky (approved):
LGTM

yash431garg (approved):
LGTM... Can you add disable state in close text when opening is processing?

PR #1058, opened 2025-03-20
fix: update env var name reference.

samott:
Fixes a typo in the env var name from PROD-909.

### DIFF
--- a/lib/env-vars.ts
+++ b/lib/env-vars.ts
@@ -19,5 +19,5 @@ export const getBonkOneTimeLimit = () => {
   if (!process.env.BONK_ONE_TIME_LIMIT) {
     return 3_000_000;
   }
-  return Number(process.env.ONE_TIME_LIMIT);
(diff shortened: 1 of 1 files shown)


marvinmarnold (approved):

lib/env-vars.ts
```diff
@@ -19,5 +19,5 @@ export const getBonkOneTimeLimit = () => {
   if (!process.env.BONK_ONE_TIME_LIMIT) {
     return 3_000_000;
   }
-  return Number(process.env.ONE_TIME_LIMIT);
+  return Number(process.env.BONK_ONE_TIME_LIMIT);
```
yash431garg, line 22: Please update env example as well accordingly.

yash431garg (changes requested):

lib/env-vars.ts
samott: It's already correct in the env file.

PR #1060, opened 2025-03-21
PROD-936: limit one time credits sent during mb open

samott:
Adds a limit (per mystery box) on the amount of credits awarded to a user.

The AC specified 200 as the default limit but we've already had campaign boxes with > 200, so I went with 1000.

This PR also has a fix for the test for the BONK limit.


psypersky (approved):

yash431garg:
Please add the env in the repo for testing. Also FYI, there is some breaking changes in PROD-894.

PR #1061, opened 2025-03-24
Prod 944: Server side randomization of second order

yash431garg:
Random option constraint on second order selection. If user haven't selected the give random option the code will throw an error.


prisma/schema.prisma
```diff
@@ -158,6 +158,7 @@ model QuestionAnswer {
   questionOption   QuestionOption @relation(fields: [questionOptionId], references: [id])
   userId           String
   user             User           @relation(fields: [userId], references: [id])
+  isRandomOption   Boolean        @default(false)
```
samott, line 161: Please add a comment (here and in Notion).

marvinmarnold: `isAssigned2ndOrderOption` could be more explicit

actions/answers/addRandomOption.ts
marvinmarnold, line 1: We aren't supposed to add new server actions.

app/hooks/useRandom.ts
```diff
@@ -16,7 +16,9 @@ export function useRandom({ min, max }: RandomProps) {
 
   const generateRandom = useCallback(
```
marvinmarnold, line 17: The randomness can't happen on the client. This needs to be decided on the server.

actions/answers/addRandomOption.ts
```diff
@@ -0,0 +1,28 @@
+"use server";
+
+import { getJwtPayload } from "@/app/actions/jwt";
+import prisma from "@/app/services/prisma";
+
+export async function addRandomOption(random: number | undefined) {
```
marvinmarnold, line 6: 1. This is letting the frontend decide what the random option is. This function should accept a questionId and then find all the user's QuestionAnswers and randomly assign one server side.
2. `random` is a bad variable name. If this variable were to stay, it should be `randomQuestionOptionId`. But based on #1, it should be replaced with `questionId`.

app/components/Deck/Deck.tsx
```diff
@@ -201,6 +203,9 @@ export function Deck({
         setRandom(
           question.questionOptions.findIndex((option) => option.id === number),
         );
+
+        await addRandomOption(number);
```
marvinmarnold, line 207: It's hard to maintain this code if the client is responsible for deciding when to make random assignment. We should do this from the server when the user first sees the question, as part of `markQuestionAsSeenButNotAnswered`.

marvinmarnold (changes requested):

prisma/schema.prisma
yash431garg: By Notion, Are you referring to database structure document?

actions/answers/addRandomOption.ts
```diff
@@ -0,0 +1,28 @@
...
+  const payload = await getJwtPayload();
+
+  if (!payload) return;
```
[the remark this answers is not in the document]
yash431garg: This action is deleted.

actions/answers/addRandomOption.ts
yash431garg: Make sense. Just updated the codebase to handle all the randomization server side.

app/hooks/useRandom.ts
yash431garg: Make sense. Just updated the codebase to handle all the randomization server side.

actions/answers/addRandomOption.ts
yash431garg: I have removed this action

app/components/Deck/Deck.tsx
yash431garg: Make sense

prisma/schema.prisma
samott: Yes. The new name for this column is much clearer though, so I'm less concerned about the need for docs now.

psypersky:
flacky test, re-runned and worked, but I haven't actually test it

actions/answers/markQuestionAsSeenButNotAnswered.ts
```diff
@@ -2,10 +2,14 @@
...
 import { getJwtPayload } from "@/app/actions/jwt";
 import prisma from "@/app/services/prisma";
+import { getRandomInteger } from "@/app/utils/randomUtils";
 import { chargeUserCredits } from "@/lib/credits/chargeUserCredits";
 import { AnswerStatus } from "@prisma/client";
 
-export async function markQuestionAsSeenButNotAnswered(questionId: number) {
+export async function markQuestionAsSeenButNotAnswered(
+  questionId: number,
+  max: number,
```
marvinmarnold, line 11: numOptions would be a more clear variable name.

But really, the client should not be responsible for passing in this value. Otherwise they could, for example, supply max=1 for all questions. This would let exploiters be able to have all their bots always do 2nd order for A.

app/hooks/useRandom.ts
```diff
@@ -16,7 +16,9 @@ export function useRandom({ min, max }: RandomProps) {
...
     ({ min, max }: RandomProps) => {
-      setRandom(getRandomInteger(min, max));
+      const randomValue = getRandomInteger(min, max);
```
marvinmarnold, line 19: What is this function still being used for?

marvinmarnold (changes requested):

app/hooks/useRandom.ts
yash431garg: This isn't used anywhere right now. I kept it thinking that may be we use it in future. But removed now to maintain code quality.

actions/answers/markQuestionAsSeenButNotAnswered.ts
yash431garg: Thanks for the suggestion. I was leaning more toward not updating too much code as It was one of the important part of our flow. But based on your comment It make sense.

__tests__/actions/answer.test.ts
```diff
@@ -134,11 +138,50 @@ describe("Validate points logs for completing questions and decks", () => {
...
+        percentageGiven: 50,
+        percentageGivenForAnswerId: questionOptions[1].id,
+        timeToAnswerInMiliseconds: 3638,
+        deckId: deckId,
+      });
+    } catch (error: any) {
+      expect(error.message).toBe(
+        `User with id: ${userId} second order respose doesn't match the give random option id for question id ${deckQuestionId}.`,
+      );
+    }
```
samott, line 154: This test will fail (false positive) if answerQuestion doesn't throw.

actions/answers/markQuestionAsSeenButNotAnswered.ts
```diff
@@ -13,20 +14,37 @@ export async function markQuestionAsSeenButNotAnswered(questionId: number) {
   const userId = payload.sub;
   const questionOptions = await prisma.questionOption.findMany({
     where: { questionId },
+    include: {
```
marvinmarnold, line 17: I think the current logic would allow isAssigned2ndOrderOption to be overwritten if invoked multiple times. The user would have to pay credits each time, but they could alter the value of the field.

app/components/Deck/Deck.tsx
```diff
@@ -100,26 +99,18 @@ export function Deck({
...
-  const { random, generateRandom, setRandom } = useRandom({
-    min,
-    max,
-  });
+
   const { start, reset, getTimePassedSinceStart } = useStopwatch();
   const [isTimeOutPopUpVisible, setIsTimeOutPopUpVisible] = useState(false);
   const [numberOfAnsweredQuestions, setNumberOfAnsweredQuestions] = useState(0);
   const [isCreditsLow, setIsCreditsLow] = useState(false);
+  const [random, setRandom] = useState(0);
```
marvinmarnold, line 107: Can't this random variable be remove entirely?

__tests__/actions/answer.test.ts
```diff
@@ -104,7 +108,7 @@ describe("Validate points logs for completing questions and decks", () => {
     });
   });
 
-  it("should allow a user to answer a question once", async () => {
+  it("should not allow to answer a question if no random option selected", async () => {
```
marvinmarnold, line 111: There should be a separate test for mc and binary since they have different expected behavior.

marvinmarnold (changes requested):

app/components/Deck/Deck.tsx
yash431garg: No, we are using it.

__tests__/actions/answer.test.ts
yash431garg: Sure

actions/answers/markQuestionAsSeenButNotAnswered.ts
yash431garg: I will double check you this. But I think we do createMany and it should fail because of unique constraint for the same question id.

yash431garg: I just checked they cannot alter the value because of unique constraint.

app/components/Deck/Deck.tsx
yash431garg: We are just using this to show in the ui what we have saved in backend.

actions/answers/markQuestionAsSeenButNotAnswered.ts
```diff
@@ -13,20 +14,37 @@ export async function markQuestionAsSeenButNotAnswered(questionId: number) {
...
+      isAssigned2ndOrderOption:
+        index === random &&
+        questionOptions[0].question.type === QuestionType.MultiChoice,
       selected: false,
     }));
     await prisma.questionAnswer.createMany({
       data: answerData,
     });
+    return {
+      random:
```
marvinmarnold, line 43: Random is unclear variable name. `assigned2ndOrderOption`?

Returning 0 for binary is a bit confusing. Null or undefined would be more explicit. At least add some comments.

samott (approved):

psypersky (approved):
I think a comment explaining why we assign our server value to a random variable would be nice, but not blocking it, can merge.

PR #1062, opened 2025-03-24
PROD-914: Don't query backend if parameters are unavailable

samott:
During the first render of the page, the URL parameters are (sometimes) unavailable, which means we send a request to the backend with an undefined question ID, triggering a Prisma error.

Returning a promise triggers suspense (i.e. the Chompy loading screen) further up the tree. There is a better solution to this problem, but we'd need NextJS 15.x.

### DIFF
--- a/app/application/answer/reveal/[questionId]/page.tsx
+++ b/app/application/answer/reveal/[questionId]/page.tsx
@@ -49,6 +49,9 @@ const NotAvailableYet = ({ msg }: { msg: string }) => {
 };
 
 const RevealAnswerPage = async ({ params }: Props) => {
+  // Parameters can be undefined on the first render;
+  // return a promise to trigger suspense further up
+  // the tree until we have the values.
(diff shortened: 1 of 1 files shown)


psypersky (approved):

psypersky (approved):
is it done that way because of suspense?

samott:
> is it done that way because of suspense?

Yes... so we can keep the loader visible until the parameters are available.

app/application/answer/reveal/[questionId]/page.tsx
```diff
@@ -49,6 +49,9 @@ const NotAvailableYet = ({ msg }: { msg: string }) => {
 };
 
 const RevealAnswerPage = async ({ params }: Props) => {
+  if (params.questionId === undefined)
```
Comment (marvinmarnold) [inline app/application/answer/reveal/[questionId]/page.tsx:52]: Please add a comment so future devs aren't confused.

PR #1063, opened 2025-03-24
PROD-953: Fix answer calculation for MC questions.

samott:
Only selected answers should be included in data sent to the mechanism engine.

Fixes the first order (correct answer) calculation but doesn't touch the second order calculation as I don't have a way of knowing if the results are correct while the dashboard is also incorrect.

### DIFF
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -157,16 +157,13 @@ const calculateMultiChoiceCorrectAnswer = async (questionIds: number[]) => {
     const questionOptions = question.questionOptions;
     const correctOption = questionOptions.find((option) => option.isCorrect);
 
-    if (!!correctOption) {
-      correctOptionIds.push(correctOption.id);
-      continue;
-    }
-
     const optionsList = questionOptions.map((option) => option.id);
 
     const body = {
       first_order_answers: questionOptions.flatMap((option) =>
-        option.questionAnswers.filter(qa => qa.selected).map(() => optionsList.indexOf(option.id)),
+        option.questionAnswers
+          .filter((qa) => qa.selected)
+          .map(() => optionsList.indexOf(option.id)),
       ),
       second_order_answers: questionOptions.map((option) =>
         option.questionAnswers
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -155,7 +155,6 @@ const calculateMultiChoiceCorrectAnswer = async (questionIds: number[]) => {
 
   for (const question of questions) {
     const questionOptions = question.questionOptions;
-    const correctOption = questionOptions.find((option) => option.isCorrect);
 
     const optionsList = questionOptions.map((option) => option.id);
 
--- a/app/utils/algo.ts
+++ b/app/utils/algo.ts
@@ -166,7 +166,7 @@ const calculateMultiChoiceCorrectAnswer = async (questionIds: number[]) => {
 
     const body = {
       first_order_answers: questionOptions.flatMap((option) =>
-        option.questionAnswers.map(() => optionsList.indexOf(option.id)),
+        option.questionAnswers.filter(qa => qa.selected).map(() => optionsList.indexOf(option.id)),
       ),
       second_order_answers: questionOptions.map((option) =>
         option.questionAnswers


marvinmarnold (approved):

marvinmarnold (changes requested):
@mleiwe should review this

mleiwe (commented):
Looks good to me. As I'm not too familiar with typescript let me just explain what I think the change is doing.
"Instead of just getting all the options, now it only considers the options which have been selected by the user."

My concern is what happens when there is an option not selected by any user as a first order, but has second order estimates. Would the mapping handle this? Apologies for the noob question
@yash431garg, @samott 

Example `questionId=4776` and `questionOptionId=11368`, option `11370` should win according to my tests

SQL query can be found here
```
SELECT "qo"."questionId", "qo"."id", "Question"."type", count(CASE WHEN "QuestionAnswer"."selected" = 't' THEN 1 END) as "user_votes", avg("QuestionAnswer"."percentage") as "averageSecondOrder" 
FROM "QuestionOption" as qo
JOIN "QuestionAnswer" ON "QuestionAnswer"."questionOptionId" = "qo"."id"
JOIN "Question" ON "Question".id = "qo"."questionId"
GROUP BY "qo"."questionId", "qo"."id", "Question"."type"
HAVING count(CASE WHEN "QuestionAnswer"."selected" = 't' THEN 1 END) = 0
```

samott:
> Looks good to me. As I'm not too familiar with typescript let me just explain what I think the change is doing. "Instead of just getting all the options, now it only considers the options which have been selected by the user."
> 
> My concern is what happens when there is an option not selected by any user as a first order, but has second order estimates. Would the mapping handle this? Apologies for the noob question. I think it should but it might be worth checking. @yash431garg, @samott
> 
> Example `questionId=4776` (`questionOptionId=11368` is not selected by anyone), option `11370` should win according to my tests
> 
> SQL query (to find example cases) can be found here
> 
> ```
> SELECT "qo"."questionId", "qo"."id", "Question"."type", count(CASE WHEN "QuestionAnswer"."selected" = 't' THEN 1 END) as "user_votes", avg("QuestionAnswer"."percentage") as "averageSecondOrder" 
> FROM "QuestionOption" as qo
> JOIN "QuestionAnswer" ON "QuestionAnswer"."questionOptionId" = "qo"."id"
> JOIN "Question" ON "Question".id = "qo"."questionId"
> GROUP BY "qo"."questionId", "qo"."id", "Question"."type"
> HAVING count(CASE WHEN "QuestionAnswer"."selected" = 't' THEN 1 END) = 0
> ```

The change here only affects the first order answers being submitted. I think the second order results are probably not right.

Do you think you could provide an example question or two from the database (maybe including one for the scenario you mentioned above) along with the expected input to the mechanism engine endpoint?

yash431garg:
I am not sure. Is this bug was introduced after some changes?

mleiwe:
@samott for question 4776
My mapping of `optionId`s
```
mapping = {11368:0, 11369:1, 11370:2, 11371:3}
```

The values I got from python
```
First orders from qid 4776: 
[1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3]

Second orders from qid 4776:
 [
[5.0, 27.0, 24.0, 0.0, 50.0, 89.0, 58.0, 100.0, 14.0, 0.0, 97.0, 50.0, 0.0, 0.0, 100.0, 0.0, 0.0], 
[25.0, 50.0, 0.0, 50.0, 17.0, 36.0, 0.0, 28.0, 50.0, 50.0, 26.0, 50.0, 61.0, 50.0], 
[100.0, 100.0, 100.0, 77.0, 50.0, 50.0, 97.0, 96.0, 100.0, 87.0, 50.0, 100.0, 81.0], 
[100.0, 50.0, 50.0, 31.0, 75.0, 0.0, 56.0, 50.0, 50.0, 7.0, 68.0, 50.0, 18.0, 44.0, 50.0, 81.0, 100.0, 50.0, 32.0, 1.0]
]
```

samott:
> @samott for question 4776 My mapping of `optionId`s
> 
> ```
> mapping = {11368:0, 11369:1, 11370:2, 11371:3}
> ```
> 
> The values I got from python
> 
> ```
> First orders from qid 4776: 
> [1, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 3]
> 
> Second orders from qid 4776:
>  [
> [5.0, 27.0, 24.0, 0.0, 50.0, 89.0, 58.0, 100.0, 14.0, 0.0, 97.0, 50.0, 0.0, 0.0, 100.0, 0.0, 0.0], 
> [25.0, 50.0, 0.0, 50.0, 17.0, 36.0, 0.0, 28.0, 50.0, 50.0, 26.0, 50.0, 61.0, 50.0], 
> [100.0, 100.0, 100.0, 77.0, 50.0, 50.0, 97.0, 96.0, 100.0, 87.0, 50.0, 100.0, 81.0], 
> [100.0, 50.0, 50.0, 31.0, 75.0, 0.0, 56.0, 50.0, 50.0, 7.0, 68.0, 50.0, 18.0, 44.0, 50.0, 81.0, 100.0, 50.0, 32.0, 1.0]
> ]
> ```

So a question like this was being skipped by the PWA and never made it to the mechanism engine. I've updated the code on our side to ensure the mechanism engine is still consulted.

I also verified that the arguments being sent for that question match what you gave above, so I suppose the second order calculations are good after all.

marvinmarnold (approved):
Why was a question like this being skipped?

samott:
> Why was a question like this being skipped?

Maybe it made sense when we only had objective MC?

PR #1065, opened 2025-03-24
PROD-963: Make credit cost an input field rather than select box.

samott:
Make the credit cost control in the admin dashboard an input field so any amount can be entered.


psypersky (approved):
LGTM

yash431garg (approved):
LGTM

PR #1072, opened 2025-03-27
fix: wrap DialogTitle with Dialog

yash431garg:


marvinmarnold:
Works well.

But I wonder if the version upgrade was really needed in the first place? Maybe it would be safer to revert that upgrade so we don't need to touch code. Is there a chance that we've missed some other pieces of code that will break other parts of the site?

yash431garg:
> Works well.
> 
> But I wonder if the version upgrade was really needed in the first place? Maybe it would be safer to revert that upgrade so we don't need to touch code. Is there a chance that we've missed some other pieces of code that will break other parts of the site?

I am not sure if version upgrade is really needed but I have checked we are using the updated package for this purpose only. 

@samott It seems that you have updated this package. Is there any specific reason?

samott:
> > Works well.
> > But I wonder if the version upgrade was really needed in the first place? Maybe it would be safer to revert that upgrade so we don't need to touch code. Is there a chance that we've missed some other pieces of code that will break other parts of the site?
> 
> I am not sure if version upgrade is really needed but I have checked we are using the updated package for this purpose only.
> 
> @samott It seems that you have updated this package. Is there any specific reason?

I don't remember upgrading the package - maybe it was Neel?

samott:
> > Works well.
> > But I wonder if the version upgrade was really needed in the first place? Maybe it would be safer to revert that upgrade so we don't need to touch code. Is there a chance that we've missed some other pieces of code that will break other parts of the site?
> 
> I am not sure if version upgrade is really needed but I have checked we are using the updated package for this purpose only.
> 
> @samott It seems that you have updated this package. Is there any specific reason?

So we didn't actually have the dialog component before. Neel added it. I guess the current version of that component brought changes that affect components that depend on the dialog building blocks.

I don't know if it's worth trawling through old versions of dialog to find one that doesn't need the changes in this PR.

In any case, we should get this merged because the breakage is happening in other branches.

samott (approved):

yash431garg:
@marvinmarnold based on Shaun comment should we merge this or you want me to do version downgrade?

marvinmarnold (approved):
Sounds like momentum is to continue with the upgraded version and adopt this PR.

PR #1075, opened 2025-03-28
PROD-947 fix overflow in rewards chip

psypersky:

Fix overflow on rewards button when on small screen and update design (icons and text) as per design:

https://linear.app/gator/issue/PROD-947/handle-overflow-case-for-rewards-chip-in-deckinfo-card

UPDATE: The "Total n Cards" text not show's to the right of the buttons when there's no image as per Yihwan instructions


app/components/PreviewDeckCard/PreviewDeckCard.tsx
```diff
@@ -111,13 +120,31 @@ const PreviewDeckCard = ({
...
+                  <div className="ml-1">
+                    <TrophyQuestionMarkIcon width={13} height={14} />
+                  </div>
+                  <div className="flex flex-col ml-2">
+                    {deckCreditCost > 0 ? (
+                      <>
+                        <span className="opacity-50 pr-1 text-left text-xs">
+                          Rewards up to{" "}
+                        </span>
+                        <span className="text-left text-xs">
```
samott, line 132: I think this also needs `font-bold`.

psypersky: yup, fixed

samott (dismissed):

marvinmarnold (commented):
Not sure if the card count was intended to be in scope. But it's different than the designs.
Screenshot 2025-03-30 at 22 10 15

psypersky:
@marvinmarnold Yihwan said text should go to the right when there's no image. It's updated

marvinmarnold:
Doesn't look like text went to the expected location.

Screenshot 2025-04-01 at 09 51 56

marvinmarnold (approved):

PR #1079, opened 2025-03-30
PROD-964 and PROD-965: list community questions on admin panel and add-to-list action

samott:
PROD-964: Implement ask inbox on admin panel.

PROD-965: Implement assignment to community deck feature.

For PROD-965, some of the complexity from the ticket wasn't needed in the end:

- We can easily identify the draft deck as it belongs to the community stack (since it is added there immediately after creation).
- The `addedToDeckAt` field is available as `DeckQuestion.createdAt`.

I have added a `specialId` field which gives us a way to identify special decks such as the community ask deck; we have the flexibility to add others in the future.


marvinmarnold (dismissed):

psypersky (approved):
LGTM

PR #1081, opened 2025-03-31
wip: ask questions users

yash431garg:


prisma/schema.prisma
```diff
@@ -70,6 +70,7 @@ model Question {
   chompResults                ChompResult[]
   stack                       Stack?                        @relation(fields: [stackId], references: [id])
   stackId                     Int?
+  isSubmittedByUser           Boolean                       @default(false)
```
samott, line 73: We need to record which user submitted the question also.

samott (commented):
Are you confident these questions won't show up anywhere in the system?

app/schemas/ask.ts
```diff
@@ -0,0 +1,59 @@
...
+    }, `Max image size allowed is ${IMAGE_UPLOAD_SIZE_STRINGS.MEDIUM}.`)
+    .refine((files) => {
+      if (files && files.length > 0) {
+        return IMAGE_VALID_TYPES.includes(files[0].type);
+      }
+      return true;
+    }, "Only .jpg, .jpeg, .png and .webp formats are supported."),
+  questionOptions: z
+    .object({
+      id: z.number().optional(),
```
samott, line 36: I think the user shouldn't be providing the ID.

yash431garg: User isn't providing the id it is an optional schema field

samott: Is it filtered out if it is included in the request - doesn't look like it?

__tests__/actions/ask/question.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    await createAskQuestion(mockQuestionData);
+
+    const result = await prisma.question.findFirst({
+      where: {
+        isSubmittedByUser: true,
+      },
+      include: {
+        questionOptions: true,
+      },
+    });
```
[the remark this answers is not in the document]
yash431garg: Yes, but wasn't able to thought of other way of doing this.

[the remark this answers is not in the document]
samott: You could at least use `orderBy` to get the most recent one.

app/schemas/ask.ts
yash431garg: I removed these from the scheme.

app/actions/ask/question.ts
marvinmarnold, line 1: No new server actions.

Although I could be convinced this is an exception. I believe the proper use of server actions is for form data submission.

app/actions/ask/question.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (!payload) return null;
+  const validatedFields = askQuestionSchema.safeParse(data);
+
+  if (!validatedFields.success) {
+    return { errorMessage: formatErrorsToString(validatedFields) };
+  }
+
+  const questionData = {
+    ...validatedFields.data,
```
marvinmarnold, line 25: It would be safer to explicitly parse all of fields that we want from validatedFields. Otherwise, it's possible to sneak in other values. For example, revealAtToken or stackId could currently get set by a malicious user.

It would also avoid having to explicitly set some fields to undefined.

app/actions/ask/question.ts
```diff
@@ -0,0 +1,46 @@
...
+    isSubmittedByUser: true,
+    createdBy: payload.sub,
+  };
+
+  await prisma.question.create({
+    data: {
+      ...questionData,
+      questionOptions: {
+        createMany: {
+          data: validatedFields.data.questionOptions,
```
marvinmarnold, line 39: Similarly, we should explicitly parse the fields we want to use. Otherwise, a malicious user could set fields like calculatedIsCorrect and we may never notice.

prisma/schema.prisma
```diff
@@ -70,8 +71,11 @@ model Question {
...
+  createdBy                   String?
```
marvinmarnold, line 75: This name is ambiguous. Unclear if it's a userId, username, or something else.

marvinmarnold (changes requested):

samott:
I tested this and it works.

The question doesn't show up on the answer page using its id but it's because the reveal date is not set. Probably we should explicitly exclude questions with `isSubmittedByUser = true` and no parent deck.

prisma/schema.prisma
samott: I updated the name since I duplicated the schema update in my branch too.

app/actions/ask/question.ts
samott: I think we are actually ok here because zod ensures `validatedFields` only has the fields from the schema.

yash431garg:
> I tested this and it works.
> 
> The question doesn't show up on the answer page using its id but it's because the reveal date is not set. Probably we should explicitly exclude questions with `isSubmittedByUser = true` and no parent deck.

Technically we won't make it an active question until someone accept it as a part of deck. Adding what you suggested we may need to update the whole codebase even reward section. Should we go for it?

samott:
Can you look into the tests that are failing? They only seem to affect this branch.

marvinmarnold (approved):

PR #1087, opened 2025-04-04
build(PROD-1001): run tests in isolated branch

marvinmarnold:
Create isolated Neon branch for every Github branch being tested. If branch is `main`, then create a new Neon branch every time. Otherwise, reuse the same branch.

@psypersky I don't see an option to create a branch with schema only in the API: https://api-docs.neon.tech/reference/createprojectbranch

Also, after implementing everything, I realized I could have simplified things by creating the compute endpoint at the same time as the test branch. Everything is working with branch creation and endpoint creation as separate steps so leaving as-is for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,88 +9,510 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      # Step 1: Checkout the repository
-      - uses: actions/checkout@v2
+      # Step 1: Check out the repository
+      - name: Check out the repository
+        id: checkout
+        uses: actions/checkout@v2
+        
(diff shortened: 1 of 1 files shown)


psypersky:
@marvinmarnold 

looks like its a separate doc

neon branch create --schema-only

https://neon.tech/docs/guides/branching-schema-only

marvinmarnold:
> @marvinmarnold
> 
> looks like its a separate doc
> 
> neon branch create --schema-only
> 
> https://neon.tech/docs/guides/branching-schema-only

Thanks @psypersky. I'm going to close this PR and reopen on a new branch so that I can test changes.

PR #1089, opened 2025-04-04
build(PROD-1001): run tests in isolated branch

marvinmarnold:
Create isolated Neon branch for every Github branch being tested. If branch is main, then create a new Neon branch every time. Otherwise, reuse the same branch. Start with an empty DB. Schema only.

Follow-up to: https://github.com/gator-labs/chomp/pull/1087

--
Context from Slack: https://gator-labs.slack.com/archives/C06054PPNSY/p1743784638470569

The overall goal is to reduce test flakiness. One of the causes of the flakiness was that the Vercel preview link that we use for manually testing each feature branch currently uses the same DB that is used by the tests.
This would mean that tests would fail because of activity from visiting web preview. Or web preview may see weird behavior from tests running.

This PR makes it so that:
- every Github branch has a Neon DB branch created just for tests. This is in addition to the preexisting Neon DB branches for every Vercel preview
- Non-main GH branches have a single Neon DB branch created for them. Every time tests run on that branch, they'll reuse the same DB. We don't spin up new DBs every time to prevent creating too many Neon DB branches.
- The main GH branch is an exception. Out of an abundance of caution, every time tests run on main GH branch, a new Neon DB branch is created, exclusively for that run.
- DB is based off of the schema of it's parent branch, but no data is copied over. Tests can assume DB starts empty.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,88 +9,515 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      # Step 1: Checkout the repository
-      - uses: actions/checkout@v2
+      # Step 1: Check out the repository
+      - name: Check out the repository
+        id: checkout
+        uses: actions/checkout@v2
+        
+      - name: Print debug info
+        run: |
+          echo "====================== DEBUG INFO ======================"
+          echo "GitHub event name: ${{ github.event_name }}"
+          echo "GitHub event action: ${{ github.event.action }}"
+          echo "Deployment status: ${{ github.event.deployment_status.state }}"
+          echo "Deployment URL: ${{ github.event.deployment_status.target_url }}"
+          echo "Deployment environment: ${{ github.event.deployment.environment }}"
+          echo "Ref: ${{ github.event.deployment.ref }}"
+          echo "Workspace: ${{ github.workspace }}"
+          echo "Run ID: ${{ github.run_id }}"
+          echo "Run attempt: ${{ github.run_attempt }}"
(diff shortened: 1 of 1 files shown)


psypersky (approved):
I looks like either should work or not, no middle ground so LGTM

nice error management

yash431garg (approved):
Overall lgtm... Verified that there aren't any env leak

psypersky:
We are waiting to merge the release first, just to be sure :p

mleiwe (approved):

mleiwe:
Oops sorry didn't realise my review didn't get approved

mleiwe:
@marvinmarnold 
I tried replicating this for mech engine and got quite a few errors which I solved but in your create branch curl
create_json=$(curl --silent --request POST \
              --url "https://console.neon.tech/api/v2/projects/$NEON_PROJECT_ID/branches" \
              --header "accept: application/json" \
              --header "authorization: Bearer $NEON_API_KEY" \
              --header "content-type: application/json" \
              --data "{
                \"branch\": {
                  \"parent_id\": \"$parent_id\",
                  \"name\": \"$TEST_BRANCH\",
                  \"init_source\": \"schema-only\"
                },
                \"endpoints\": [
                  {
                    \"type\": \"read_write\",
                    \"autoscaling_limit_min_cu\": 0.25,
                    \"autoscaling_limit_max_cu\": 2
                  }
                ]
              }")
              
your `init_source` is set to "schema-only" which is fine but you won't copy over any data is that intended?   

cross ref: https://api-docs.neon.tech/reference/createprojectbranch

PR #1091, opened 2025-04-04
PROD-1007: Limit precision of BONK stored in the database and shown to users.

samott:
Limit precision of BONK stored in the database and shown to users.


marvinmarnold (approved):

yash431garg:
Is there a way to test this in preview?

PR #1092, opened 2025-04-05
Admin panel: hard-coded cost per question limit.

samott:
Admin panel: hard-coded cost per question limit.

### DIFF
--- a/app/schemas/deck.ts
+++ b/app/schemas/deck.ts
@@ -114,7 +114,7 @@ export const deckSchema = z
     revealTokenAmount: z.number().min(0),
     revealAtDate: z.date({ message: "Reveal at date is required" }),
     revealAtAnswerCount: z.number().min(0).nullish(),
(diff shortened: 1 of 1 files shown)


app/schemas/deck.ts
```diff
@@ -114,7 +114,7 @@ export const deckSchema = z
     revealTokenAmount: z.number().min(0),
     revealAtDate: z.date({ message: "Reveal at date is required" }),
     revealAtAnswerCount: z.number().min(0).nullish(),
-    creditCostPerQuestion: z.number().min(0).nullish(),
+    creditCostPerQuestion: z.number().min(0).max(100).nullish(),
```
marvinmarnold, line 117: Can this be an env var that defaults to 100? Don't want to have to push code change if this value needs to be adjusted.

yash431garg (approved):

psypersky (approved):
LGTM

PR #1093, opened 2025-04-07
PROD-925: use new mech endpoints for binarymc answer calculations

samott:
- Don't calculate answers in PWA.
- Use new rewards views on answer page.
- Update rewards views to not break when used in Prisma queries.


psypersky:
@samott does this has things that we can test in the UI ?

PR #1095, opened 2025-04-08
PROD-967 create a script to identify bonk owed to user

psypersky:
Note: This is the linear issue https://linear.app/gator/issue/PROD-967/fix-historical-bonk-owed
PROD-962 Is the parent Isue


package.json
```diff
@@ -90,6 +95,7 @@
     "jsonwebtoken": "8.5.1",
     "jwks-rsa": "^3.1.0",
     "link-preview-js": "^3.0.12",
+    "log-update": "4",
```
marvinmarnold, line 98: Should this be a dev dependency?

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
marvinmarnold, line 1: It's `prizes` not `prices`

scripts/mystery-box/mystery-box-price-update-tx.ts
```diff
@@ -0,0 +1,139 @@
...
+      isValidSolanaTxHash = false;
+    }
+
+    if (!isValidSolanaTxHash) {
+      writeRowWithError(row, "INVALID_SOLANA_TX_HASH");
+      continue;
+    }
+
+    try {
+      await prisma.chainTx.upsert({
```
marvinmarnold, line 95: should all the db updates for a row happen in a single db transaction?

scripts/mystery-box/mystery-box-price-update-tx.ts
```diff
@@ -0,0 +1,139 @@
...
+        where: { id: mbpId },
+        data: { claimHash: newHash },
+      });
+    } catch (err) {
+      console.error(err);
+      writeRowWithError(row, "DB_MBP_UPDATE_ERROR");
+      continue;
+    }
+
+    appendFileSync(
```
marvinmarnold, line 132: Doesn't the MysteryBox also need to have it's status updated?

scripts/mystery-box/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+1. Run the mystery-box-find-unsuccessful-prices.ts script to generate a CSV file containing all problematic transactions. Note that this script may take several hours to complete execution.
+
+2. Once the CSV file is generated, we need to clean it for false positives. For some reason, the Solana RPC API falsely returns that a tx does not exist sometimes, even 5 times in a row! So you need to run and re-run this file until the transaction count of your file stops changing; in my case it was three times.
+
+❗Make sure to fill the variables FILE_NAME and IS_RE_CLEAN with the correct information.
+
+3. (Optional) If you want to separate your file by error type for better handling, you can use results-clean-false-positives.ts file to generate a new file with your filtered MBP.
+
+4. (Optional) You can run results-add-dates.ts to add dates to MBP. TODO: Might be better to edit the script to add them from the beginning.
```
marvinmarnold, line 13: Not clear what it means to add dates

scripts/mystery-box/README.md
```diff
@@ -0,0 +1,26 @@
+# Mystery Box Script
+
+Follow these steps to identify problematic transactions:
+
+1. Run the mystery-box-find-unsuccessful-prices.ts script to generate a CSV file containing all problematic transactions. Note that this script may take several hours to complete execution.
+
+2. Once the CSV file is generated, we need to clean it for false positives. For some reason, the Solana RPC API falsely returns that a tx does not exist sometimes, even 5 times in a row! So you need to run and re-run this file until the transaction count of your file stops changing; in my case it was three times.
+
+❗Make sure to fill the variables FILE_NAME and IS_RE_CLEAN with the correct information.
```
marvinmarnold, line 9: Are these environment variables? Unclear where they need to be set. What is a reclean? The same thing as a rerun?

scripts/mystery-box/README.md
```diff
@@ -0,0 +1,26 @@
...
+3. (Optional) If you want to separate your file by error type for better handling, you can use results-clean-false-positives.ts file to generate a new file with your filtered MBP.
+
+4. (Optional) You can run results-add-dates.ts to add dates to MBP. TODO: Might be better to edit the script to add them from the beginning.
+
+5. Create each transaction manually to pay users and at the end of each record add the TX id to update the DB with `mystery-box-price-update-tx.ts`. That's it, your users are happier now!
+
+## Commands
+
+```bash
+yarn dev:mystery-box-prices-inspect-txs
```
marvinmarnold, line 20: This is a lot of steps. Ideally, there would only be two scripts. One to identify all the MBP where BONK is owed. Another to fix the entries. The granular functionality of these existing scripts can be abstracted.

scripts/mystery-box/mystery-box-price-update-tx.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    // === Step 2 Point MBP to new ChainTX
+    let updatedMbp: Prisma.MysteryBoxPrizeGetPayload<{
+      select: { claimHash: true };
+    }>;
+
+    try {
+      updatedMbp = await prisma.mysteryBoxPrize.update({
+        where: { id: mbpId },
+        data: { claimHash: newHash },
```
marvinmarnold, line 124: Don't additional fields on MBP need to be updated: status, claimedAt, claimChainTxId

marvinmarnold: Note finalizedAt is sometimes null currently and sometimes has a date. I'm curious why.
Screenshot 2025-04-18 at 19 01 03

scripts/mystery-box/mystery-box-price-update-tx.ts
```diff
@@ -0,0 +1,139 @@
...
+  const fileTxt = readFileSync(FILE_PATH, { encoding: "utf8" });
+  const fileTxtRows = fileTxt.split(/\r?\n/);
+
+  function writeRowWithError(row: string, error: string) {
+    appendFileSync(cleanedFilePath, row + `,${error},\n`);
+  }
+
+  // Process the data rows (skip the header)
+  for (let i = 0; i < fileTxtRows.length; i++) {
+    const row = fileTxtRows[i];
```
marvinmarnold, line 38: Don't need to change. But in the future, could break up these long functions into smaller ones so it's easier to read.

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
```diff
@@ -0,0 +1,408 @@
...
+  await prisma.$transaction([
+    prisma.$executeRaw`
+        DO $$
+        BEGIN
+          IF NOT EXISTS (
+            SELECT 1 FROM pg_indexes 
+            WHERE tablename = 'MysteryBoxPrize' 
+            AND indexname = 'idx_mysteryboxprize_claimhash'
+          ) THEN
+            CREATE INDEX idx_mysteryboxprize_claimhash ON "MysteryBoxPrize" ("claimHash");
```
marvinmarnold, line 79: Are these indices being permanently created?  This is surprising behavior.

scripts/mystery-box/README.md
```diff
@@ -0,0 +1,26 @@
+# Mystery Box Script
+
+Follow these steps to identify problematic transactions:
+
+1. Run the mystery-box-find-unsuccessful-prices.ts script to generate a CSV file containing all problematic transactions. Note that this script may take several hours to complete execution.
+
+2. Once the CSV file is generated, we need to clean it for false positives. For some reason, the Solana RPC API falsely returns that a tx does not exist sometimes, even 5 times in a row! So you need to run and re-run this file until the transaction count of your file stops changing; in my case it was three times.
```
marvinmarnold, line 7: Did you confirm that it still fails if you run single threaded without all the parallel worker stuff?

scripts/mystery-box/mystery-box-price-update-tx.ts
psypersky: All MysteryBoxes are opened and the system things they were successful, I don't think so

scripts/mystery-box/mystery-box-price-update-tx.ts
psypersky: I guess we could do it yeah

scripts/mystery-box/README.md
psypersky: Yeah I guess this was more for me to test and research, I'll make everything only two steps

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
psypersky: No, the issue mentions we will use replica for reading so this indices will be deleted every hour right?

scripts/mystery-box/README.md
psypersky: Yes, the clean script is a single thread/loop with enough wait time between calls and that happens quite often

psypersky: And at the end I ended up making everything "single threaded" because of that

scripts/mystery-box/mystery-box-price-update-tx.ts
psypersky: @marvinmarnold Yes I noted the null in finalizedAt, maybe we can create an issue for that? I have no idea how deep that rabbit whole goes

psypersky: Quote: Don't additional fields on MBP need to be updated: status, claimedAt, claimChainTxId, finalizedAt.
 
 claimChainTxId doesn't exist in postgres, just claimHash which is already updated
 
 as for the rest it was actually claimed when it was originally claimed and I guess we would say it was really finalized now, so the finalizedAt I guess if we really wanna be exact

scripts/mystery-box/README.md
psypersky: No its variables at the top of the file, I'll be more specific

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
psypersky: Altho, in order to be sure I'll make them temporary anyway

scripts/mystery-box/mystery-box-price-update-tx.ts
marvinmarnold: How do we know that all the MysteryBoxes are in the right status? It looks like we are only making our query  based on MBP state: https://github.com/gator-labs/chomp/pull/1095/files#diff-04951141fd58db4d6d85bf969041014c32ba6a43faf07441254087c80831fda0R140

Screenshot 2025-04-19 at 16 48 13

scripts/mystery-box/mystery-box-price-update-tx.ts
marvinmarnold: Gotcha, yes lets set finalizedAt also then too.

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
marvinmarnold: Gotcha. I added a note in https://linear.app/gator/issue/PROD-721/audit-and-fix-db-settings to add these indexes to prod. Do you think that's something we should do? Could it be a cause of slowness for PWA running in prod?

scripts/mystery-box/README.md
marvinmarnold: Cool. Once the PR is final, I'll try running myself and see if I can replicate.

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
psypersky: I'm not sure if we would need them on prod since this script will always be running on replica and when the indexes are deleted it will just create them again. I think for now they are fine just being re-created every time they are deleted, I don't feel strongly about any of both options. But yeah we can review it on that issue that's a good idea

scripts/mystery-box/README.md
psypersky: @marvinmarnold as for the cleaning step (check which not existing TX actually not exist) would you rather have an extra step for cleaning false positives or that the first step cleans it false negatives automagically ?

psypersky: I would say lets keep it as an extra step for now and when automating it we can just run it after the first step, that way also we wont forget that about that "problem" and hide it over too much magic, but tell me what ya think

marvinmarnold: Sure, we can do an extra step. Let's just make sure the exact commands and steps to run things is clear.

scripts/mystery-box/mystery-box-find-unsuccessful-prices.ts
marvinmarnold: Do you suspect that queries in prod might also be made much faster if these indices are added? Do our queries in prod not use these columns often?

psypersky: Yes, I'll link this convo in PROD-721 this queries are very heavy and we might need them on prod eventually or even now

scripts/mystery-box/mystery-box-price-update-tx.ts
psypersky: @marvinmarnold 

Ok, so this is interesting, we have 24 claimed  MysteryBoxPrizes with their MysteryBoxes not 'Opened' and all from the same user in the same day.

I think we can safely ignore the MysteryBox status since it looks like only a bug would cause a MysteryBox not opened with a claimed MysteryBoxPrize, do you agree on this? 

And on the other hand I think we need to make an issue about those 24 claimed MBP with not opened MB, here's the query.

```sql
SELECT 
    mbp."id" AS prize_id,
    mbp."status" AS prize_status,
    mbp."prizeType" AS prize_type,
    mbp."amount",
    mbp."claimedAt",
    mbp."tokenAddress",
    mbp."claimHash",
    mbt."id" AS trigger_id,
    mbt."triggerType" AS trigger_type,
    mb."id" AS box_id,
    mb."status" AS box_status,
    mb."userId",
    u."username",
    u."telegramUsername"
FROM 
    "MysteryBoxPrize" mbp
JOIN 
    "MysteryBoxTrigger" mbt ON mbp."mysteryBoxTriggerId" = mbt."id"
JOIN 
    "MysteryBox" mb ON mbt."mysteryBoxId" = mb."id"
LEFT JOIN
    "User" u ON mb."userId" = u."id"
WHERE 
    mbp."status" = 'Claimed'
    AND mb."status" != 'Opened'
ORDER BY
    mbp."claimedAt" DESC,
    mb."userId";
```

scripts/mystery-box/README.md
psypersky: @marvinmarnold fixed as discussed, there are only 3 steps now

psypersky:
@marvinmarnold refactores and tested over the weekend, works OK, updated the missing info, deleted unnecessary steps, single threaded everything and merged updates in transactions. 

Also added a variable so we can choose a date to which start looking faulty MBP, that way we only need to look for the not analyzed ones by date :)

scripts/mystery-box/mystery-box-price-update-tx.ts
marvinmarnold: Great. Yes please create a follow up ticket for these 24.

scripts/mystery-box/mystery-box-find-unsuccessful-prizes.ts
```diff
@@ -0,0 +1,458 @@
...
+      continue;
+    }
+
+    let newPrizes: Array<MysteryBoxPrizeWithChainTX>;
+    try {
+      newPrizes = await prisma.$queryRawUnsafe<
+        Array<MysteryBoxPrizeWithChainTX>
+      >(`
+        SELECT 
+          mbp."id" as "mbpId",
```
marvinmarnold, line 159: This means that we will query the same tx hash multiple times if there are many prizes that were claimed with the same hash. Is that correct?

scripts/mystery-box/mystery-box-find-unsuccessful-prizes.ts
```diff
@@ -0,0 +1,458 @@
...
+    try {
+      debugFilter(`[Filter W${index}] getting TX`);
+
+      tx = await CONNECTION.getParsedTransaction(mbp.mbpClaimHash, {
+        commitment: "confirmed",
+        maxSupportedTransactionVersion: 0,
+      });
+
+      // Transaction does not exists at all on Blockchain (problematic!)
+      if (tx === null) {
```
marvinmarnold, line 257: I still think it makes more sense to do a retry directly, instead of having a separate claen-false-positives step

marvinmarnold (approved):
Looks like it's working. I don't love that:

1) Same tx hash is requested multiple times
2) Not retrying getParsedTransaction if it returns null.
3) All the noise from the worker stuff

Here is a ChatGPT refactor, but we can use what's in your PR.

```
import { PrismaClient } from "@prisma/client";
import { ParsedTransactionWithMeta } from "@solana/web3.js";
import debug from "debug";
import fs from "fs";
import path from "node:path";

// Project helper that exposes a configured Solana RPC connection
import { CONNECTION } from "../../app/utils/solana";

/**
 * Mystery‑Box Prize → Solana reconciliation (simplified)
 * -----------------------------------------------------
 * • single‑threaded, no batching/worker queues
 * • every claimHash is fetched from chain **once** and cached
 * • retries up to **MAX_RETRIES** times when the transaction is not yet available
 * • any mystery‑box prize with a missing or failed TX is written to a CSV
 * • logs every TX processed ⇒ hash / ok / error / running count
 */

type ISODateString = string;
const START_AT_CREATED_AT: ISODateString | null = null; // put ISO date to limit history or leave null

const OUTPUT_PATH = path.join(
  __dirname,
  `results-${new Date().toISOString().replace(/[:.]/g, "-")}.csv`,
);

// Debug helpers --------------------------------------------------------------
const log = debug("mbp");
debug.enable("mbp");

// DB ------------------------------------------------------------------------
const prisma = new PrismaClient();

// Config --------------------------------------------------------------------
const MAX_RETRIES = 10; // ← retry this many times when tx === null
const RETRY_DELAY_MS = 1_000; // wait between retries when tx is still null

// Types ---------------------------------------------------------------------
interface MBP {
  mbpId: string;
  mbpClaimHash: string;
  ctFinalizedAt: string | null;
  mbUserId: string;
  chainTxTokenAmount: number;
  chainTxRecipientAddress: string;
}

type ErrorType = "NOT_EXISTS" | "FAILED";
interface CSVRow extends MBP {
  error: ErrorType;
}

// Helpers -------------------------------------------------------------------
function isTransactionSuccessful(tx: ParsedTransactionWithMeta): boolean {
  return !(tx.meta?.err);
}

function sleep(ms: number) {
  return new Promise((resolve) => setTimeout(resolve, ms));
}

async function fetchTxWithRetry(hash: string): Promise<ParsedTransactionWithMeta | null> {
  for (let attempt = 1; attempt <= MAX_RETRIES; attempt++) {
    const tx = await CONNECTION.getParsedTransaction(hash, {
      commitment: "confirmed",
      maxSupportedTransactionVersion: 0,
    });

    if (tx !== null) {
      return tx; // success
    }

    if (attempt < MAX_RETRIES) {
      log(`🔄  Retry ${attempt}/${MAX_RETRIES - 1} — TX still null for ${hash}`);
      await sleep(RETRY_DELAY_MS);
    }
  }
  return null; // still null after retries
}

async function fetchClaimedTokenPrizes(): Promise<MBP[]> {
  const startDateSql = START_AT_CREATED_AT
    ? `'${new Date(START_AT_CREATED_AT).toISOString()}'`
    : "'1970-01-01'";

  return prisma.$queryRawUnsafe<MBP[]>(`
    SELECT
      mbp."id"                  AS "mbpId",
      mbp."claimHash"           AS "mbpClaimHash",
      ct."finalizedAt"          AS "ctFinalizedAt",
      mb."userId"               AS "mbUserId",
      ct."tokenAmount"          AS "chainTxTokenAmount",
      ct."recipientAddress"     AS "chainTxRecipientAddress"
    FROM "MysteryBoxPrize" mbp
    JOIN "MysteryBoxTrigger" mbt ON mbp."mysteryBoxTriggerId" = mbt."id"
    JOIN "MysteryBox"        mb  ON mbt."mysteryBoxId"        = mb."id"
    LEFT JOIN "ChainTx"       ct  ON mbp."claimHash"          = ct."hash"
    WHERE mbp."prizeType" = 'Token'
      AND mbp."status"    = 'Claimed'
      AND mbp."claimHash"  IS NOT NULL
      AND mbp."createdAt" >= ${startDateSql}
    ORDER BY mbp."claimHash", mbp."createdAt" DESC;
  `);
}

// Main ----------------------------------------------------------------------
async function main(): Promise<void> {
  log("⏳ Fetching claimed‑token prizes from DB …");
  const prizes = await fetchClaimedTokenPrizes();
  log(`🎁 Retrieved ${prizes.length} rows`);

  const txCache = new Map<string, { ok: boolean; error?: ErrorType }>();
  const csvRows: CSVRow[] = [];

  let txCount = 0; // running counter for logs

  for (const prize of prizes) {
    const hash = prize.mbpClaimHash;
    if (!hash) continue; // safety guard

    // cache miss ⇒ hit RPC (with retry logic)
    if (!txCache.has(hash)) {
      let ok = false;
      let error: ErrorType | undefined;

      try {
        const tx = await fetchTxWithRetry(hash);

        if (tx === null) {
          error = "NOT_EXISTS";
        } else if (!isTransactionSuccessful(tx)) {
          error = "FAILED";
        } else {
          ok = true;
        }
      } catch (rpcErr) {
        log(`⚠️  RPC error while fetching ${hash}:`, rpcErr);
        error = "NOT_EXISTS";
      }

      txCache.set(hash, { ok, error });
    }

    const cacheEntry = txCache.get(hash)!;

    // increment + log every tx seen (cached or newly fetched)
    txCount++;
    log(`TX #${txCount} | hash=${hash} | ok=${cacheEntry.ok} | error=${cacheEntry.error ?? "none"}`);

    if (!cacheEntry.ok && cacheEntry.error) {
      csvRows.push({ ...prize, error: cacheEntry.error });
    }
  }

  // CSV output --------------------------------------------------------------
  fs.writeFileSync(
    OUTPUT_PATH,
    [
      [
        "mbpId",
        "mbpClaimHash",
        "error",
        "ctFinalizedAt",
        "mbUserId",
        "chainTxTokenAmount",
        "chainTxRecipientAddress",
      ].join(","),
      ...csvRows.map((r) =>
        [
          r.mbpId,
          r.mbpClaimHash,
          r.error,
          r.ctFinalizedAt ?? "",
          r.mbUserId,
          r.chainTxTokenAmount?.toString() ?? "",
          r.chainTxRecipientAddress ?? "",
        ].join(","),
      ),
    ].join("\n"),
  );

  log(`✅ Done. ${csvRows.length} problematic prizes written to ${OUTPUT_PATH}`);
  process.exit(0);
}

main().catch((e) => {
  console.error(e);
  process.exit(1);
});
```

scripts/mystery-box/mystery-box-find-unsuccessful-prizes.ts
psypersky: yup, also when fixing it we could put the same tx hash for multiple rows in our file

scripts/mystery-box/mystery-box-find-unsuccessful-prizes.ts
psypersky: I guess we could but there is no definitive number of tries that we have in order to be sure that it definitively doesn't exist, what I do right now if clean the file again and again until the number of rows doesn't change, If I hardcode a number of tries there we could have either false posives in our rows or taking waaay to long because I harcoded a big number of tries, its a bit tricky :p

psypersky: ooooor, maybe I could run script 2 inside script 1 until row number stop changing :D

psypersky:
@marvinmarnold agree, once we use the first script as "single thread" a lot of the complexity doesn't make sense, we can simplify it by a lot like in the GPT refactor, but testing these scripts take a while so

1- would it be fine to leave the simplification for when we implement this https://linear.app/gator/issue/PROD-968/fix-steady-state-bonk-owed ?
2 - I'm not sure about the hardcoded retry number becausfe from my test sometimes the RCP just throws a bunch of nulls in a row like some times "decides to not work", its weird, so if we do that it might still leak false positives
3 - Same tx hash requested multiple times

If you agree we can address all these improvements on 968

marvinmarnold:
> If you agree we can address all these improvements on 968

Yup, sounds good.

samott (approved):
I'll approve based on the previous comments and the fact that Marvin says it works.

The code is quite complex and I'm not sure I understand the file naming system. But from a high level it looks ok.

PR #1097, opened 2025-04-08
PROD-990: Do not infer 2nd order for binary

samott:
- Do not infer 2nd order answers.
- Add tests.
- Fix existing answer tests.

### DIFF
--- a/__tests__/actions/answer-binary.test.ts
+++ b/__tests__/actions/answer-binary.test.ts
@@ -138,5 +138,7 @@ describe("answer binary question", () => {
       },
     });
     expect(questionAnswer).toHaveLength(2); // We expect 4 answers because we created 4 options
+    expect(questionAnswer[0].percentage).toBeNull();
+    expect(questionAnswer[1].percentage).toBe(50);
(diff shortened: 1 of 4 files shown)


yash431garg (approved):
LGTM... we also need to make sure that mech-engine is processing based on updated answer now.

samott:
> LGTM... we also need to make sure that mech-engine is processing based on updated answer now.

@mleiwe Anything needed on the mechanism engine side to support this? We are sending 0 for the non-selected entries (and in the database percentage will be NULL)

mleiwe:
@samott and @marvinmarnold 
Relevant code for SP binary is [here](https://github.com/gator-labs/mechanism-engine/blob/03421a2bd5ecef13472f78ae61fe55f9a40fbd6f/chomp/answers.py#L55). Complexity for this change is small, will need to refactor the tests though which might take a bit of time.

samott:
@mleiwe Sorry, I just did some testing and in fact the null values are not sent to the mechanism engine after all - they are just skipped.

PR #1098, opened 2025-04-09
PROD-1013, PROD-1014, PROD-1015: new answer page: header, question and chart

samott:
New answer page header and queries, layout, hook.

And: question component.

And: chart.

And: feature flag.


yash431garg:
@samott I am getting this error. If I add optional chaining operator I get NaN
Screenshot 2025-04-10 at 11 17 43 AM
Screenshot 2025-04-10 at 11 24 13 AM

samott:
> @samott I am getting this error. If I add optional chaining operator I get NaN Screenshot 2025-04-10 at 11 17 43 AM Screenshot 2025-04-10 at 11 24 13 AM

Ok, I'll fix. I don't think we have a design for the case where user's haven't opened the box yet.

yash431garg:
Font size and padding looks off to me

samott:
I added some tests @yash431garg.

lib/answerStats/getStats.ts
```diff
@@ -0,0 +1,203 @@
...
+      rewardStatus: "no-reward",
+    };
+  }
+
+  const answerSelected = userAnswers.find((ua) => ua.selected);
+  const isFirstOrderCorrect =
+    correctAnswer?.id === answerSelected?.questionOptionId;
+  const isPracticeQuestion = question.creditCostPerQuestion === 0;
+
+  const isQuestionAnsweredByUser = userAnswers.length > 0;
```
yash431garg, line 142: This will consider question as answered even when it timeout/Viewed. But we cannot show user result because he haven't selected anything.

samott: Nice catch. Fixed.

marvinmarnold:
This question renders as expected

Screenshot 2025-04-14 at 14 42 32

But this one is missing the rewards summary underneath the question title. I think it's missing because it's a practice question. https://chomp-7yakiooqn-gator-labs.vercel.app/application/answer/reveal-new/6644
Screenshot 2025-04-14 at 14 42 58

psypersky (approved):
LGTM

samott:
> This question renders as expected
>
> But this one is missing the rewards summary underneath the question title. I think it's missing because it's a practice question.

It's because it's a legacy question with no chomp result. I maybe think it's better to show 0/0 in this case, so I've updated the logic.

samott:
> > This question renders as expected
> > But this one is missing the rewards summary underneath the question title. I think it's missing because it's a practice question.
> 
> It's because it's a legacy question with no chomp result. I think it's better to show 0/0 in this case, so I've updated the logic.

I actually reverted this change: we hide the header if the user didn't answer a non-legacy question, so I think it should be the same for legacy questions.

Legacy questions with a chomp result will show the header with the rewarded tokens.

PR #1099, opened 2025-04-09
Prod 1014:new answer page best answer

yash431garg:


app/application/answer/reveal-new/[questionId]/page.tsx
```diff
@@ -24,15 +28,49 @@ const RevealAnswerPageNew = ({ params }: Props) => {
 
   const stats = result.data.stats;
 
+  const isBinary = stats.type === QuestionType.BinaryQuestion;
+
+  const answerSelected = stats.userAnswers.find((ua) => ua.selected);
+
+  let answerContent = <></>;
```
Comment (samott) [inline app/application/answer/reveal-new/[questionId]/page.tsx:35]: Why not use a ternary comparison here and make this const?

app/application/answer/reveal-new/[questionId]/page.tsx
```diff
@@ -24,15 +28,49 @@ const RevealAnswerPageNew = ({ params }: Props) => {
...
+
+  if (!!isBinary) {
+    answerContent = (
+      <>
```
Comment (samott) [inline app/application/answer/reveal-new/[questionId]/page.tsx:39]: Is this fragment wrapper needed?

components/QuestionPreviewCard/QuestionPreviewCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+              </Modal>
+            </>
+          )}
+        </div>
+      </div>
+    </div>
+  );
+}
+
+export default QuestionPreviewCard;
```
samott, line 70: I also created this component for the ask feature. We can use yours after the merge, but on the ask page the image button is still shown when there is no image (with label "No Image") - we should check if it should be the same design here.

yash431garg: Screenshot 2025-04-11 at 10 20 25 AM
It is mentioned in figma to not show 'View Image' If there aren't any image.

app/application/answer/reveal-new/[questionId]/page.tsx
Comment (yash431garg) [reply app/application/answer/reveal-new/[questionId]/page.tsx:39]: Yes, It will have more components in the follow up PR.

components/QuestionPreviewCard/QuestionPreviewCard.tsx
samott: Ok, makes sense.

psypersky (approved):
LGMT

PR #1100, opened 2025-04-10
PROD-957: Ask hi-fi design

samott:
Implement new ask design.

- Hi-fi design for frontend.
- Improved admin panel.
- Reward credits for asking and getting answers.


prisma/schema.prisma
```diff
@@ -184,6 +186,17 @@ model QuestionAnswer {
   @@index([userId, selected])
 }
 
+model AskQuestionAnswer {
```
marvinmarnold, line 189: Can you add comments to the schema about what this table is for? Notion as well.

I'm surprised this table is necessary and that we can't just use a regular QuestionAnswer instead.

samott: The reason for adding the new table is that we want a unique entry, which isn't possible with `QuestionAnswer`, since we have two or four of them.

samott: Done.

marvinmarnold:
Not able to submit MC question
https://github.com/user-attachments/assets/9e4183fc-72dc-4a0b-8ea8-fe9d4bdad0ff

samott:
> Not able to submit question https://github.com/user-attachments/assets/9e4183fc-72dc-4a0b-8ea8-fe9d4bdad0ff

Fixed.

PR #1102, opened 2025-04-12
Apply ask rewards using a DB trigger.

samott:
Switch to triggers for ask rewards.

I created a table `AskQuestionAnswer` in the main ask branch to track answer / user ID combos in the FATL as it avoids a giant mess with multiple relations between tables.

Not totally convinced this is better than the non-trigger approach, but don't have a strong opinion.

### DIFF
--- a/actions/answers/answerQuestion.ts
+++ b/actions/answers/answerQuestion.ts
@@ -198,37 +198,6 @@ export async function answerQuestion(request: SaveQuestionRequest) {
         );
       }
 
-      if (
-        questionOptions[0].question.isSubmittedByUser &&
-        questionOptions[0].question.createdByUserId
-      ) {
-        const CREDITS_REWARD = Number(
-          process.env.NEXT_PUBLIC_ASK_ANSWERED_CREDITS_REWARD ?? 0,
-        );
-
-        const aqa = await tx.askQuestionAnswer.create({
-          data: {
-            questionId: request.questionId,
-            userId,
-          },
-        });
-
-        // first and second order required to be eligible for a reward
-        if (CREDITS_REWARD && request.percentageGiven !== undefined) {
-          fungibleAssetRevealTasks.push(
-            tx.fungibleAssetTransactionLog.createMany({
-              data: {
-                type: TransactionLogType.AskQuestionAnswered,
-                askQuestionAnswerId: aqa.id,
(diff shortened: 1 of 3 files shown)


marvinmarnold (commented):
What are the pros/cons of the two approaches? 

Test cases would def be helpful here.

samott:
> What are the pros/cons of the two approaches?

In-code approach:
- Seems simpler to me. If it breaks it's easier to debug whereas any faults in the SQL code tend to manifest in weird opaque ways, esp. with Prisma involved.
- Can share the configured payout values rather than duplicate/maintain separately in the SQL script.

Trigger approach:
- Data integrity - we're guaranteed to have the right matching sets of records.
- No risk of forgetting to update any new code path with the credits logic.
But:
- No real way of checking the query at compile time; we have to rely on tests to validate the SQL.

> Test cases would def be helpful here.

I can add more tests. We do have some coverage from the existing PR.

The trigger approach is pretty neat. I'm warming to it.

marvinmarnold (approved):

samott:
> > > What are the pros/cons of the two approaches?
> > 
> > 
> > In-code approach:
> > 
> > * Seems simpler to me. If it breaks it's easier to debug whereas any faults in the SQL code tend to manifest in weird opaque ways, esp. with Prisma involved.
> > * Can share the configured payout values rather than duplicate/maintain separately in the SQL script.
> > * More flexible: e.g. if we get a request to update the logic to do some weird stuff before deciding whether to apply the reward in the future (query an API or something), it's easy to do that in code.
> > 
> > Trigger approach:
> > 
> > * Data integrity - we're guaranteed to have the right matching sets of records.
> > * No risk of forgetting to update any new code path with the credits logic.
> >   But:
> > * No real way of checking the query at compile time; we have to rely on tests to validate the SQL.
> > 
> > > Test cases would def be helpful here.
> > 
> > 
> > I can add more tests. We do have some coverage from the existing PR.
> > The trigger approach is pretty neat. I'm warming to it.
> 
> Yup, I agree with your analysis. I don't feel strongly that we use triggers for this. But I do think it's a useful tool that we can incorporate more generally. Defer to you on final decision.

I've done the work, so let's go with it. We can explore using this pattern more for other stuff.

I'll follow up with more tests during in the next cycle.

PR #1108, opened 2025-04-16
feat: charge user credit after second order submission

yash431garg:
Updated logic to charge user credit after second order submission instead of question first seen.


psypersky (dismissed):
looks good and tested

marvinmarnold (approved):

PR #1109, opened 2025-04-17
Prod 1030: Answer page fix

yash431garg:


samott:
Looks good.

I personally prefer to keep component props with components since they are almost always only used with the component itself, but I don't feel too strongly about it.

samott (approved):

psypersky (approved):
looks good and same thing Shaun said about the component props

PR #1111, opened 2025-04-17
PROD-1032: Close ask image preview on click outside or esc

samott:
- Close modal with click outside of image or press Escape.
- Limit max height of modal so that the X button is always visible.


psypersky (dismissed):
works

yash431garg (commented):
I am not sure If we have to show it full screen in desktop. But preview isn't looking good in desktop full screen. 
Screenshot 2025-04-18 at 10 56 05 AM

Also, Can you make it a utility because we need to implement this feature in new answer page as well.

samott:
> Also, can you make it a utility because we need to implement this feature in new answer page as well?

In what sense?

PR #1118, opened 2025-04-21
PROD-1043: Allow logged out users to see daily decks.

samott:
Allow logged out users to see daily decks.


psypersky (dismissed):
@samott makes sense, should we create an issue to update tests maybe?

samott:
> @samott makes sense, should we create an issue to update tests maybe?

Probably the right thing to do is to update them in this PR. :-)

marvinmarnold (approved):

PR #1122, opened 2025-04-23
Add tests got getStack and improve test generator

psypersky:
Made an attempt to make our tests simpler by having a tool which makes it very simple to create and connect Stacks, Decks, Questions, QuestionsOptions, Answers and Users, a big percentage of the tests is simply creating and deleting data, a lot of the times the errors are in the tests instead of the main code, so here's the IDontWannaWriteMoreTestsV2.0 👯 

Also:
Follow up to changes in https://github.com/gator-labs/chomp/pull/1047
Verify the totalRewardAmount and creditCost
check answeredQuestions in case of MCQ and Binary


marvinmarnold (commented):
Is the failure in `get-next-deck-id-query` related to these changes?

PR #1125, opened 2025-04-23
PROD-1045 make mystery box campaigns hideable

psypersky:
Should be possible to hide campaigns from Mystery Box page. It will not be possible to claim MBP for campaigns that have been hidden.

Un/hide will be done by a dev, directly accessing DB.

### DIFF
--- a/lib/mysteryBox/getCampaigns.ts
+++ b/lib/mysteryBox/getCampaigns.ts
@@ -4,15 +4,15 @@ import { EMysteryBoxStatus } from "@prisma/client";
 import "server-only";
 
 /**
- * @description Get new and unclaimed campaigns allowed for user
- * @returns {Array<Object>} An array containing a single object with item details
- * @returns {string} returns[0].id - Unique identifier for the item
- * @returns {string} returns[0].name - Name of the item
- * @returns {string} returns[0].infoTitle - Title of the item's information
- * @returns {string} returns[0].infoBody - Body text of the item's information
- * @returns {boolean} returns[0].isEligible - Eligibility status of the item
+ * Get new and unclaimed campaigns allowed for user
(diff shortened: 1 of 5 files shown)


prisma/schema.prisma
```diff
@@ -379,6 +379,7 @@ model CampaignMysteryBox {
   CampaignMysteryBoxAllowlist CampaignMysteryBoxAllowlist[]
   MysteryBoxTrigger           MysteryBoxTrigger[]
   boxType                     String
+  enabled                     Boolean                       @default(true)
```
samott, line 382: Would have preferred `isEnabled` here.

psypersky: yeah that's more booleany, I'll update it

lib/mysteryBox/getCampaigns.ts
```diff
@@ -4,15 +4,15 @@ import { EMysteryBoxStatus } from "@prisma/client";
 import "server-only";
 
 /**
- * @description Get new and unclaimed campaigns allowed for user
- * @returns {Array<Object>} An array containing a single object with item details
```
marvinmarnold, line 8: Why remove all these comments?

psypersky: removed the old JsDoc commend and replace them with new Typescript Types :D, just leaved the regular comments

yash431garg (commented):
I think it is possible to invoke createCampaignMysteryBox for disabled campaigns because it doesn't respect the enabled flag.

yash431garg (changes requested):
I think it is possible to invoke createCampaignMysteryBox for disabled campaigns because it doesn't respect the enabled flag

PR #1126, opened 2025-04-23
add logging for error when marking questions as seen

psypersky:
More visibility for error when marking errors as seen, I have some suspicions about the behavior of questions when this error happens, this is as part of the investigation for 1039 skipping of questions, nobody is able to reproduce the bug still but some users say it happens


samott (approved):

app/components/Deck/Deck.tsx
```diff
@@ -150,19 +151,47 @@ export function Deck({
...
+        if (response?.hasError) {
+          Sentry.captureMessage(
+            `Error calling markQuestionAsSeenButNotAnswered()`,
+            {
+              level: "error",
+              tags: {
+                category: "deck-errors",
+              },
+              extra: {
+                questionId: question.id,
```
marvinmarnold, line 170: Add userId?

marvinmarnold (approved):

PR #1130, opened 2025-04-24
PROD-1028: ensure consistent UI when 2nd order response missing

samott:
- Questions without 2nd order response are seen as "viewed".
- New states in query added for unseen and incomplete (no second order) to differentiate between cases.
- Question status query complexity has been offloaded into a view.


yash431garg:
@samott looks like some checks are failing

psypersky (approved):
I'm not getting the 500

PR #1139, opened 2025-04-30
feat(query): update to avoid pulling all answers records for stacks w…

yash431garg:


samott (approved):

psypersky (approved):
looks good and tested

PR #1141, opened 2025-05-01
PROD-1039: Automatic question skipping

samott:
I've been unable to reproduce the auto-skipping behaviour locally or on prod, but from the logs it seems like the action is being called multiple times for one question, triggering an exception and the skipped-question phenomenon. Possibly ultimately due to a weird dependency issue with the hook on the frontend.

For now I've updated the action to quietly do nothing if called with the same question, which I think makes sense anyway.


yash431garg (approved):
I am not sure either about the exact reason.

PR #1144, opened 2025-05-06
PROD-968 fix-steady-state-bonk-owed

psypersky:
Run the script to detect problematic MysteryBoxPrizes daily and report it to Slack, created an api and a vercel cron

also 

refactored the scripts/mystery-box/mystery-box-find-unsuccessful-prizes.ts to be simpler and "single thread"

### DIFF
--- a/app/api/cron/detect-bonk-owned/route.ts
+++ b/app/api/cron/detect-bonk-owned/route.ts
@@ -1,4 +1,4 @@
-import { main as findUnsuccessfulPrices } from "@/scripts/mystery-box/mystery-box-find-unsuccessful-prizes";
+import { main as findUnsuccessfulPrizes } from "@/scripts/mystery-box/mystery-box-find-unsuccessful-prizes";
 import { main as resultsCleanFalsePositives } from "@/scripts/mystery-box/results-clean-false-positives";
 import * as Sentry from "@sentry/nextjs";
 import dayjs from "dayjs";
@@ -21,7 +21,7 @@ export async function GET(request: Request) {
     .subtract(1, "hour")
     .toDate()
     .toISOString();
-  const resFindFilePath = await findUnsuccessfulPrices(
+  const resFindFilePath = await findUnsuccessfulPrizes(
     startOfDayUTCMinusOneHour,
   );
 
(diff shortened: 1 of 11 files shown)


app/api/cron/detect-bonk-owned/route.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export async function GET(request: Request) {
+  const secret = process.env.CRON_SECRET || "";
+
+  const authHeader = request.headers.get("authorization");
+  if (authHeader !== `Bearer ${secret}`) {
+    return new Response("Unauthorized", { status: 401 });
+  }
+
+  const startOfDayUTCMinusOneHour = dayjs()
```
marvinmarnold, line 18: Will this miss prizes that fail in the last hour of the day?

marvinmarnold: No, now I see it's 1 day - 1 hour.

marvinmarnold (commented):
@samott one thing to consider about this is that it will use the prod DB and Helius keys. Ideally, we should use the prod read only replica or chomp-replica and different RPC keys for non-critical processes like this. Defer to you about if it's worth optimizing for now.

app/api/cron/detect-bonk-owned/route.ts
marvinmarnold, line 1: Should be `detect-bonk-owed`

marvinmarnold (changes requested):
1. I don't understand how the file can be retrieved if this is run on vercel. The filesystem is not available after cron runs.
2. It's not explicitly mentioned in issue, but I thought you were planning on outputting to Slack.

psypersky:
@marvinmarnold 

The file is only used while the function exists and its sent to Sentry as a CSV file which then sends it to Slack as an alert

If you go to the Slack channel there are already some example events  (see picture below)

Screenshot from 2025-05-07 20-24-39

And if you click the Sentry event you will see the instructions and can download the CSV file (see picture below)

Screenshot from 2025-05-07 20-21-53

app/api/cron/detect-bonk-owned/route.ts
psypersky: yeah, will fix

psypersky:
As for reading for the replica,we should probably check it tomorrow morning in the daily

marvinmarnold:
Ok thanks for the clarifications. Yes, lets review tomorrow.

psypersky:
@marvinmarnold updated the script to use the new variable and fixed the spell errors, alto re-rested, new alert is on slack

psypersky:
@marvinmarnold pr is ready

PR #1145, opened 2025-05-06
feat(api): add get questions v1 structure

yash431garg:


app/v1/questions/[questionId]/route.ts
```diff
@@ -0,0 +1,22 @@
...
+  { params }: { params: { questionId: string } },
+) {
+  // parse backend-secret and source headers
+  const backendSecret = request.headers.get("backend-secret");
+  if (backendSecret !== process.env.BACKEND_SECRET) {
+    return Response.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  // source corresponds to Zuplo -> Services -> API Key Service -> select consumer -> metadata -> source
+  const source = request.headers.get("source");
```
Comment (marvinmarnold) [inline app/v1/questions/[questionId]/route.ts:14]: don't need source for /get endpoint

Comment (yash431garg) [reply app/v1/questions/[questionId]/route.ts]: I think we do need  source to verify who is accessing the endpoint.

Comment (marvinmarnold) [reply app/v1/questions/[questionId]/route.ts]: I was going to say I disagree. Then I remembered there's a caveat.

Authentication: Zuplo handles authentication for us. It manages API keys and ensures that only valid API keys can access Zuplo. Then we use BACKEND_SECRET to make sure that only Zuplo can access our API. From this perspective, we don't need to do any additional checks on source.

Authorization: When there is more than one API consumer, we may want to limit the data available via API based on the source. For that, you are right that we would need the source. I think this feature is nice-to-have right now. If there's not other work to do, it could be good to support now. On the other hand, requirements may change. So it may be a waste of time to trying to scope the DB to source right now.

@samott thoughts?

Comment (samott) [reply app/v1/questions/[questionId]/route.ts]: I'd lean towards having the source check for now; we can remove it if requirements change.

app/application/answer/reveal-new/[questionId]/page.tsx
Comment (marvinmarnold) [inline app/application/answer/reveal-new/[questionId]/page.tsx:1]: Shouldn't be in this PR.

app/v1/questions/[id]/route.ts
```diff
@@ -0,0 +1,127 @@
...
+          },
+        },
+      },
+    });
+
+    // Group answers by userId
+    const groupedAnswersByUser: Record<string, QuestionAnswer[]> = {};
+
+    for (const answer of questionAnswers) {
+      if (!groupedAnswersByUser[answer.userId]) {
```
Comment (marvinmarnold) [inline app/v1/questions/[id]/route.ts:83]: I recommend refactoring into at least one additional smaller method.

marvinmarnold (changes requested):
This needs test cases right?

app/application/answer/reveal-new/[questionId]/page.tsx
Comment (yash431garg) [reply app/application/answer/reveal-new/[questionId]/page.tsx:1]: Linting changes

app/v1/questions/[id]/route.ts
Comment (yash431garg) [reply app/v1/questions/[id]/route.ts:92]: I thought about this but not sure what should I do to divide it into a smaller method.

Comment (samott) [reply app/v1/questions/[id]/route.ts:103]: That transformation could happen inside the method. In general it's good to keep routes lightweight and have the heavy lifting in a "service" type utility (see my code for example).

marvinmarnold (dismissed):

samott (dismissed):

samott (approved):

marvinmarnold (approved):

PR #1148, opened 2025-05-08
PROD-1039: automatic skipping of questions (2)

samott:
This resolves an issue which was causing the question skipping phenomenon. The problem was in the ask trigger which fails randomly if we don't explicitly set a schema prefix.

I imagine there are other instances of this issue in other raw queries throughout the app so we should definitely do an audit, but for now this one is in the answer code path and I believe responsible for the failures we've been seeing.

Edit: I went ahead and found/fixed the other instances of unprefixed tables. This PR also has some other lint fixes.


marvinmarnold (approved):
Very interesting type of error.

PR #1151, opened 2025-05-09
PROD-1085: multi question generation from admin panel

samott:
Fix test question generator.

### DIFF
--- a/app/actions/question/generateTestQuestion.ts
+++ b/app/actions/question/generateTestQuestion.ts
@@ -12,7 +12,7 @@ export async function generateBinaryTestQuestion(
   tag: string,
   creditCostPerQuestion: number | null,
   questionCount: number,
-): Promise<{ deckLink: string }> {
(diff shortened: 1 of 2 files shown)


yash431garg (approved):
Was

app/actions/question/generateTestQuestion.ts
```diff
@@ -87,37 +87,50 @@ export async function generateBinaryTestQuestion(
...
+                  : null;
+
+              await prisma.questionAnswer.create({
+                data: {
+                  userId: user.id,
+                  questionOptionId: option.id,
+                  percentage: percentage,
+                  selected: isSelectedOption,
+                  timeToAnswer: BigInt(Math.floor(Math.random() * 60000)),
+                },
```
marvinmarnold, line 122: isAssigned2ndOrderOption is not being set. I don't think that matters but just calling it out.

marvinmarnold (approved):

PR #1159, opened 2025-05-15
fix(PROD-1055): rerun tests on main without failure

marvinmarnold:
- Tests are failing on main when re-run because the branch name doesn't change. Added the runId to the branch name so it will be unique every time.
- Also added a trivial documentation change. This was just for testing tests.


yash431garg (approved):

samott (approved):
Makes sense.

PR #1163, opened 2025-05-18
feat: cache Dynamic JWKS key

marvinmarnold:
We are currently fetching a key from Dynamic's API in order to verify JWTs. Because of the nature of our serverless Vercel deployments, we are probably not getting many benefits from the current caching. 

```ts
const client = new JwksClient({
      jwksUri: jwksUrl,
      rateLimit: true,
      cache: true,
      cacheMaxEntries: 5,
      cacheMaxAge: 600000,
    });
```

Instead, this PR introduces caching with Redis. 

The hope is that these failing JWKS requests were causing our UI to be in a weird state and think that the user was not logged in for some arbitrary computations. 

Additionally context was also added to Sentry error for low credits. For debugging purposes.


marvinmarnold:
@samott @yash431garg is it expected that tests are failing now?

yash431garg (approved):
I am not sure why the tests are failing

lib/credits/chargeUserCredits.ts
```diff
@@ -30,7 +30,7 @@ export const chargeUserCredits = async (questionId: number) => {
 
   if (userTotalCreditAmount < creditCostPerQuestion) {
     throw new InsufficientCreditsError(
-      `User has insufficient credits to charge for question ${questionId}`,
+      `User ${payload.sub} has insufficient credits to charge for question ${questionId} [User amount: ${userTotalCreditAmount}, Credit cost: ${creditCostPerQuestion}]`,
```
yash431garg, line 33: We need to update this in the test as well

yash431garg (changes requested):

lib/credits/chargeUserCredits.ts
marvinmarnold: Thanks. I fixed that test.

PR #1170, opened 2025-05-19
feat: GET /questions endpoint

marvinmarnold:
- Add GET route to existing /v1/questions API route
- Only return desired subset of question info
- Also updates answer endpoint so that it accepts decimal values


marvinmarnold:
@samott for the sake of time, I'm merging this now.

PR #1174, opened 2025-05-20
fix: API fixes

marvinmarnold:
https://linear.app/gator/issue/PROD-1123/api-enforce-binary-1st-order-=-2nd-order
https://linear.app/gator/issue/PROD-1118/api-get-question-results-should-return-scores

- Restrict 2nd order option for binary
- Return answerScores and bestOption


marvinmarnold:
@samott @yash431garg merging without review so I can test e2e in staging

PR #1182, opened 2025-05-30
fix: question skipping

marvinmarnold:
Although the issue is about question skipping, impacted questions don't actually show up as skipped. Given the limited information in user reports, it's hard to be sure, but the questions seem to appear as Submitted and Viewed. 

It's also possible that https://github.com/gator-labs/chomp/pull/1148 actually fixed the issue and there has just been miscommunication about the timing of bug reports.

- Don't go to next question if problem marking as read. I'm not sure why we were ever doing that anyways. The new behavior is more user friendly. The edge case will result them in being able to see a question twice instead of getting skipped.
- Remove unnecessary params from useEffect/useCallback. I suspect this was related to the root cause of the error.
- useRef for currentQuestionIndex. Claud suggested this. 

https://linear.app/gator/issue/PROD-1130/investigate-random-question-skipping-bug-on-decks-reported-by-user

### DIFF
--- a/app/components/Deck/Deck.tsx
+++ b/app/components/Deck/Deck.tsx
@@ -96,6 +96,7 @@ export function Deck({
   const [currentQuestionIndex, setCurrentQuestionIndex] = useState(
     questions.findIndex((q) => q.status === undefined),
   );
+  const currentQuestionIndexRef = useRef(currentQuestionIndex);
 
   const [currentOptionSelected, setCurrentOptionSelected] = useState<number>();
   const [optionPercentage, setOptionPercentage] = useState(50);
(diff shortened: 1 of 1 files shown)


app/components/Deck/Deck.tsx
```diff
@@ -170,42 +194,64 @@ export function Deck({
...
 
         if (response?.random !== undefined) {
           setRandom(response.random);
         }
       } catch (error) {
+        // Only handle error if this effect hasn't been aborted
+        if (aborted) return;
+
         console.error("Error marking question as seen:", error);
-        handleNextIndex();
```
marvinmarnold, line 187: I'd also keep this change.

marvinmarnold:
Replaced by https://github.com/gator-labs/chomp/pull/1183

PR #1183, opened 2025-05-30
fix: disable accidental question hopping

marvinmarnold:
https://linear.app/gator/issue/PROD-1130/investigate-random-question-skipping-bug-on-decks-reported-by-user

We used to move to next question if problem marking question as seen. This led to a bug where we would skip questions if there was an error. We removed that, so now users may potentially see the same question again.      

Replaces https://github.com/gator-labs/chomp/pull/1182/files which had changes trying to address the root cause. But surface area of changes is too big.


samott (dismissed):
Looks good to me.

marvinmarnold:
@samott I just added user info to the Sentry message.

PR #1187, opened 2025-06-09
feat: override back logic for some pages

marvinmarnold:
https://linear.app/gator/issue/PROD-1138/create-fix-for-back-arrow-navigation-issue-in-prod-1137

- Overrides back logic for specific routes.
- Imagine a user visiting pages in the following order: A -> B -> C -> B
- When the user clicks the back button, we sometimes want to navigate to A, not C.
- But because this may not be the case for all routes, we need to be able to override the default back behavior.
- One gotcha with route matching is the dynamic part of URLs. In this case, we want to intercept `/application/decks/123`. This PR assumes that all of our IDs are integers and that we don't have integers in any other part of our URLs.

### DIFF
--- a/app/components/BackButton/BackButton.tsx
+++ b/app/components/BackButton/BackButton.tsx
@@ -1,19 +1,62 @@
+/**
+ * BackButton Component
+ * 
+ * Supports three back navigation behaviors (in priority order):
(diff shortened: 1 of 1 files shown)


samott:
Can't say I love this solution... but I guess it'll do for now.

